PR #186, opened 2014-09-03
fixed segfault occurred when Rational dump with oj_mimic_json.

kenchan:
SystemStackError occured when Rational object dump with require `oj_mimic_json`.

example here.

``` ruby
require 'active_support/core_ext'

# "{\"foo\":\"0/1\"}"
p ({foo: Rational(1/2)}).to_json

require 'oj'
require 'oj_mimic_json'

# stack level too deep (SystemStackError)
p Oj.dump({foo: Rational(1/2)})
p ({foo: Rational(1/2)}).to_json
```


shlima:
+1

Issue #234, opened 2015-05-06
version 2.1.6 can not compile in mingw32 ruby 2.1.6


Eric-Guo:
C:/Ruby21/lib/ruby/gems/2.1.0/extensions/x86-mingw32/2.1.0/oj-2.12.6/gem_make.out
```
C:/Ruby21/bin/ruby.exe -r ./siteconf20150506-2248-1oh1bxi.rb extconf.rb
>>>>> Creating Makefile for ruby version 2.1.6 on i386-mingw32 <<<<<
creating Makefile

make "DESTDIR=" clean

make "DESTDIR="
generating oj-i386-mingw32.def
compiling cache8.c
compiling circarray.c
compiling compat.c
compiling dump.c
dump.c: In function 'dump_time':
dump.c:1034:72: error: expected ')' before ';' token
dump.c:1109:1: error: expected ',' or ';' before '}' token
dump.c:1112:1: error: invalid storage class for function 'dump_ruby_time'
dump.c:1119:1: error: invalid storage class for function 'dump_xml_time'
dump.c:1219:1: error: invalid storage class for function 'dump_data_strict'
dump.c:1232:1: error: invalid storage class for function 'dump_data_null'
dump.c:1245:1: error: invalid storage class for function 'dump_data_comp'
dump.c:1308:1: error: invalid storage class for function 'dump_data_obj'
dump.c:1352:1: error: invalid storage class for function 'dump_obj_comp'
dump.c:1413:1: error: invalid storage class for function 'dump_obj_obj'
dump.c:1489:1: error: invalid storage class for function 'dump_obj_attrs'
dump.c:1682:1: error: invalid storage class for function 'dump_struct_comp'
dump.c:1721:1: error: invalid storage class for function 'dump_struct_obj'
dump.c:1808:1: error: invalid storage class for function 'dump_odd'
dump.c:1891:1: error: invalid storage class for function 'raise_strict'
dump.c:1896:1: error: invalid storage class for function 'dump_val'
dump.c:2111:1: error: invalid storage class for function 'dump_chars'
dump.c:2121:1: error: invalid storage class for function 'dump_leaf_str'
dump.c:2137:1: error: invalid storage class for function 'dump_leaf_fixnum'
dump.c:2157:1: error: invalid storage class for function 'dump_leaf_float'
dump.c:2173:1: error: invalid storage class for function 'dump_leaf_array'
dump.c:2211:1: error: invalid storage class for function 'dump_leaf_hash'
[...]

PR #235, opened 2015-05-06
Just fix missing one ).

Eric-Guo:
fix issue #234


ohler55:
You did fix the problem with this pull request and I thank you for pointing out the problem. Unfortunately you should have removed a paren instead of adding one. I made the change and checked in the removal of a paren. Please see if it works for you now.

Eric-Guo:
yes, thanks.

PR #240, opened 2015-05-29
Simplify code in Oj::Bag class

deepj:


ohler55:
Looks reasonable. Have you run tests with 1.8.7?

deepj:
Unfortunately no. I can't install this version through `ruby-build`. I'm waiting on the results from Travis.

deepj:
OK. It failed on 1.8.7. Unfortunately not due to my changes.

ohler55:
Thats okay. There are some issues with 1.8.7 and rails.

I'll look it over in more detail later and merge.

lib/oj/bag.rb
```diff
@@ -37,9 +36,9 @@ def respond_to?(m)
     # @raise [NoMethodError] if the instance variable is not defined.
     def method_missing(m, *args, &block)
       raise ArgumentError.new("wrong number of arguments (#{args.size} for 0) to method #{m}") unless args.nil? or args.empty?
-      at_m = ('@' + m.to_s).to_sym
-      raise NoMethodError.new("undefined method #{m}", m) unless instance_variable_defined?(at_m)
-      instance_variable_get(at_m)
+      result = instance_variable_get(:"@#{m}")
+      raise NoMethodError.new("undefined method #{m}", m) unless result
+      result
     end
```
ohler55: Problem here. The reason the instance_variable_defined? is called is that is $VERBOSE is true you get errors displayed on the console. I often run in that mode to catch errors in code when I am developing. I presume others do to to catch mistakes. The old was is better in this case. I'm not sure if '@' + m.to_s is more efficient than "@#{m}". Probably doesn't make much of a difference. Either is okay with me.

lib/oj/bag.rb
```diff
@@ -50,13 +49,11 @@ def eql?(other)
       ova = other.instance_variables
       iv = instance_variables
       return false if ova.size != iv.size
-      iv.each do |vid|
-        return false if instance_variable_get(vid) != other.instance_variable_get(vid)
-      end
+      return false if iv.any? { |vid| instance_variable_get(vid) != other.instance_variable_get(vid) }
       true
```
ohler55: Right idea. How about one step further.

return !iv.any? { |vid| instance_variable_get(vid) != other.instance_variable_get(vid) }

ohler55:
Make the changes and I will merge.

lib/oj/bag.rb
deepj: OK. I'd make the change with `instance_variable_defined?`. I didn't realize this use case

To another one problem. See this benchmark:

```
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]

                fast    99.381k i/100ms
                slow    97.281k i/100ms
-------------------------------------------------
                fast      3.020M (±12.1%) i/s -     14.907M
                slow      2.811M (±10.0%) i/s -     13.911M

Comparison:
                fast:  3019754.8 i/s
                slow:  2811132.8 i/s - 1.07x slower
```

Code:
```ruby
require 'benchmark/ips'

variable = 'count'

Benchmark.ips do |x|
  x.report('fast') { :"@#{variable}"}
  x.report('slow') { ('@' + variable).to_sym }
  x.compare!
end
```

Here are some useful tips how some approaches of same things are slow/fast: https://github.com/JuanitoFatas/fast-ruby

Maybe are there some useful tips how to improve C code according to it (I don't understand C code so I may be totally wrong). It depends what equivalents in C of that methods called.

lib/oj/bag.rb
deepj: I'd like to avoid this. For me it's too cryptographic. But if you want to still change it. I'll do that.

ohler55: how about 
```
return not iv.any? { |vid| instance_variable_get(vid) != other.instance_variable_get(vid) }
```
or
```
return ! iv.any? { |vid| instance_variable_get(vid) != other.instance_variable_get(vid) }
```

ohler55:
Thanks for the link on the benchmarks. Good to know. They are close but your way is a hair faster.

lib/oj/bag.rb
deepj: I think it's better to use `all?` method then. http://ruby-doc.org/core-1.8.7/Enumerable.html#method-i-all-3F

Agreed?

deepj:
Yes, it's a hair faster but any (even a bit)  better performance still counts. Ruby isn't so fast language.

lib/oj/bag.rb
ohler55: Agree!

ohler55:
No its not. Nice to use though.

deepj:
Done :) I hope everything is now okay :)

ohler55:
Very nice, thanks.

Issue #249, opened 2015-07-21
Oj.load raises SyntaxError instead of Oj::ParseError


fxfilmxf:
It looks like a recent release started raising a SyntaxError instead of an Oj::ParseError on some kinds of bad input. Is there a reason for this change?

Example:

Previous Behavior (version 2.10.4):
`Oj.load('somejunk') => # Oj::ParseError`
`Oj.load('1234otherjunk') => # Oj::ParseError`

New Behavior (version 2.12.10):
`Oj.load('somejunk') => # Oj::ParseError`
`Oj.load('1234otherjunk') => # SyntaxError`

croeck:
+1

ohler55:
You can verify that it does what you want in the github version. I wanted to try and address one more issue before releasing.

ohler55:
Release made.

croeck:
Perfect, thanks a lot!

PR #250, opened 2015-07-21
use Oj::ParseError instead of SyntaxError if no callback and multiplejsons in string

fxfilmxf:
Related to issue #249


stim371:
:+1: raising a SyntaxError currently breaks code that is expecting a ParseError to be raised if the string can't be parsed.

ohler55:
The changes make sense. I'll merge and take a look at the travis errors before releasing.

ohler55:
My mistake. I should have rejected the pull request. You checked in a binary. That was what was breaking the travis build. Fixed now. Something to keep an eye on in the future though.

fxfilmxf:
Sorry about that, I missed the binary in the PR (it's hard to see). Thanks for spending the time to figure that out.

PR #251, opened 2015-07-23
cleanup generated oj.so in build_test script, add to .gitignore

fxfilmxf:
Since I made a mistake and committed oj.so in my last pull request, it seems like it should be added to .gitignore as well as cleaned up after the build_test script runs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ Makefile
 doc
 .yardoc
 Gemfile.lock
+*.so


.gitignore
```diff
@@ -13,3 +13,4 @@ Makefile
 doc
 .yardoc
 Gemfile.lock
+ext/oj/oj.so
```
ohler55: *.so would be better

build_test.sh
```diff
@@ -39,4 +39,6 @@ cd ../../test
 ./test_all.sh
 cd ..
 
+echo "\nRemoving test generated ext/oj/oj.so"
+rm ./ext/oj/oj.so
```
ohler55: Removing after running the tests means you can not run the test twice. This does not work for me.

fxfilmxf:
Updated after feedback from @ohler55.

ohler55:
Thanks

Issue #244, opened 2015-06-24
oj causes `as_json` to behave differently from `to_json`


nornagon:
Without oj + oj_mimic_json:

```
$ rails c
irb(main):001:0> MyModel.as_json(include:{}, only:[], methods:[])
{}
irb(main):002:0> MyModel.to_json(include:{}, only:[], methods:[])
"{}"
```

With oj + oj_mimic_json

```
$ rails c
irb(main):001:0> MyModel.as_json(include:{}, only:[], methods:[])
{}
irb(main):002:0> MyModel.to_json(include:{}, only:[], methods:[])
"{\"id\":12345,\"name\":\"aoeu\",...}"
```

`to_json` includes a bunch of information that it shouldn't be including.

I'm overriding `serializable_hash` in `MyModel`:

```
  def serializable_hash(opts = {})
    opts = DefaultJsonOpts.merge(opts || {})
    super(opts)
  end
```

ohler55:
I'm a little slow getting to this. Vacation does that sometimes. Anyway, Oj does not consider any of the arguments to to_json(). I would guess that you do not have a to_json() method defined for your class and Oj is using the default which is to include each attribute of the object in the resulting JSON. I have not attempted to mimic the extra opt argument behavior in Oj. I will consider that though if you can explain what you expect and what the arguments to to_json are used for.

ohler55:
Is this a dead issue for you?

chancedowns:
I'm currently running into the same issue. We're removing attributes if their value is nil (and doing so recursively) by overriding serializable_hash like above.

Is there a better way to remove nil values?

ohler55:
I am not a rails expert or even fluent with rails. I'll need some help figuring out what you are doing. When does serializable_hash get called?

chancedowns:
It looks like it's getting called from as_json (here's 4.1, but 4.2 is the same https://github.com/rails/rails/blob/4-1-stable/activemodel/lib/active_model/serializers/json.rb#L99 )

I'm not too familiar with the inner workings of the gem, but I'm guessing that the overwriten to_json is calling as_json, which in turn calls serializable_hash

ohler55:
I think I know what is going on. I'll have to think about the best work around. Oj replaces the JSON injected to_json method which does not check to see if there is an as_json() method. I have to check if there is any problem doing that other than a performance overhead.

ohler55:
[...]

PR #252, opened 2015-08-09
Preliminary fix for #244

asurin:
@ohler55 I've made a preliminary attempt at fixing issue #244 .  I've added a new test case (Although it might be worth merging into _test_active_mimic.rb.  I'm very open to any feedback on the C changes/style as my C's pretty rusty.  I've also made on small language fix to design.txt


ohler55:
I'll have to go over the code first. First look it seems good though. Other than the requirement for sqlite3. Can you get rid of that requirement? Travis tests need to pass as well but if you can get rid of the requirement for sqlite3 they should pass I think.

asurin:
To give a little background on the issue from a Rails-centric point of view.

Most Rails' objects serialize via. two steps.  Generally, the developer will call `to_json` to actually generate the serialized data.  Internally, `to_json` will invoke `as_json`, which returns a hash containing json-compatible (basic types, or objects that also respond to an `as_json` call) objects that it can then serialize.

It looks like the JSON gem allows arbitrary parameters to be passed to `to_json`, which should then be passed through to `as_json` intact.  Oj was always invoking `as_json` without the params passed to `to_json` when it replaced it.  This patch should ensure the parameters are passed through to the internal `as_json` methods

asurin:
Yup, 100% - I'll remove that now!

ohler55:
you are a star, thanks.

asurin:
@ohler55 hmmm..it looks like ActiveRecord requires a DB adapter,a dn because this is testing the rails integration it'll break if I remove AR.  I copied the require from _test_active.rb - I'm guessing the underscore is so minitest ignores it?  (I'll confess I've mainly used rspec in the past).  I've renamed that test with an underscore too, in the hopes that it'll stop that test from being included, but I'm also 100% good with just deleting the test entirely

ohler55:
Maybe call to_json with arguments and make sure they are used in the object to_json or as_json.

asurin:
Ah good idea!  I'll get that coded up this evening or early tomorrow and added to the PR.  That'll let me slim the test down a lot - thanks for the advice!

asurin:
@ohler55 this test should now not depend on Rails, and instead explicitly test the passthrough from `to_json` to `as_json`

ohler55:
Great. I'll look and merge tomorrow. Thank you.

asurin:
Sounds great - thanks for your help with all this!

ohler55:
Looks good. Thanks. It will be released soon.

ohler55:
Did a little cleanup. Release will go out assuming travis is happy.

asurin:
Awesome - thanks for the help and the quick turnaround.  Really appreciate it!

ohler55:
I'd have to turn that around and thank you for getting the fix in.

asurin:
No worries at all, glad to help :)

chancedowns:
:+1:

Issue #255, opened 2015-08-25
oj.so built in ext looks like it's being copied to wrong directory in lib


mkillianey:
When the oj gem is built on a Linux machine, the ext/oj/oj.so file is incorrectly copied to lib/oj.so, instead of lib/oj/oj.so, and so it isn't being up correctly.  This looks to be because the Makefile is being created in "oj" instead of "oj/oj", which seems to be what [the RubyGems "gems with extensions"](http://guides.rubygems.org/gems-with-extensions/) page suggests.

The gem seems to be compensating by adding 'ext' to the require_paths, but this shouldn't be necessary, if I'm reading [the RubyGems specification](http://guides.rubygems.org/specification-reference/#require_paths=) correctly.

ohler55:
With your changes in the current branch can this be closed? I will be making a release as soon as I get confirmation 2 other bugs are closed.

mkillianey:
Absolutely...thanks!

ohler55:
Your change is release 2.12.13. I think this can be closed.

PR #256, opened 2015-08-25
Build makefile in oj/oj as per example in rubygems

mkillianey:
Fix for issue #255 

Build makefile in oj/oj as per example in rubygems: http://guides.rubygems.org/gems-with-extensions/


ohler55:
I will have to do some checking before I merge this. This was the standard when the project started. Time change though so I will check the latest docs. Thanks for the links. I'll get to it this week.

PR #268, opened 2015-10-22
Adding hard-to-find option documenation to readme

krainboltgreene:
I had an issue this morning due to the default nature of the escape encoding, and had to look at C code to find out what I could do about it. This documentation should definitely be top level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Oj gem
 
+[![Build Status](https://secure.travis-ci.org/ohler55/oj.png?branch=master)](http://travis-ci.org/ohler55/oj)
+
 A fast JSON parser and Object marshaller as a Ruby gem.
 
 ## Using
@@ -126,6 +128,46 @@ from a user and evaluating it is never a good idea from an unsecure source. The 
 they are not more than `String`s. Always check inputs from untrusted sources.
 
 
+**Rails**
+
+Although up until 4.1 Rails uses [multi_json](https://github.com/intridea/multi_json), an [issue in Rails](https://github.com/rails/rails/issues/9212) causes ActiveSupport to fail to make use Oj for JSON handling.
+There is a
+[gem to patch this](https://github.com/GoodLife/rails-patch-json-encode) for
+Rails 3.2 and 4.0. As of the Oj 2.6.0 release the default behavior is to not use
+the `to_json()` method unless the `:use_to_json` option is set. This provides
+another work around to the rails older and newer behavior.
+
+The latest ActiveRecord is able to work with Oj by simply using the line:
+
+``` ruby
+serialize :metadata, Oj
+```
+
+
+## Installation
+
+``` shell
+gem install oj
+```
+
+or in Bundler:
+
+``` ruby
+gem 'oj'
+```
+
+In version Rails 4.1, multi_json has been removed, and this patch is unnecessary and will no longer work.
+Instead, use the `oj_mimic_json` [gem](https://github.com/ohler55/oj_mimic_json) along with `oj` in your `Gemfile` to have Oj mimic the JSON gem and be used in its place by `ActiveSupport` JSON handling:
+
+```
+gem 'oj'
+gem 'oj_mimic_json'
+```
+
+## Documentation
+
+*Documentation*: http://www.ohler.com/oj, http://rubydoc.info/gems/oj
+
 **Alternative JSON Processing APIs**
 
 Oj offers a few alternative APIs for processing JSON. The fastest one is the `Oj::Doc` API. The `Oj::Doc` API takes a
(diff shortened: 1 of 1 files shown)


ohler55:
There generated docs. Here for example http://www.ohler.com/oj/Oj.html. Click on classes and then Oj.

I can understand the desire to have the options more prominently displayed though. Having said that I don't think they need to take over the landing page. If you would like to have your PR pulled in there will have to be a number of changes. We can go that route or I can make the changes here. You choice.

If you want to proceed then first, there is no need for 2 example of the use that are almost the same in the Readme. I'd prefer to keep the example under "Simple JSON Writing and Parsing Example". Moving that example to earlier in the Readme makes sense though as does changing the heading to Using as you did.

As for the options descriptions, trying to merge values into the default_options does not work. You have to use the line give as Oj.default_options = {:mode => :compat} approach as noted in the Description section. I you would like to describe the options under the description, not as code but as an unordered list I'd be glad to include that.

krainboltgreene:
It definitely does work. You might have looked at a previous edit where I forgot the `!`.

ohler55:
calling Oj.default_options.merge!() will return a Hash and modify it with the merge. It will not change the Oj.default_options though. That option Hash is generated when the call is made and is a copy of what is used internally. That merge call will not modify the options in Oj.

Back to the other points though. Placing the setting of options as the first section and labeling it "Using" is not the way I want to introduce Oj. The "Using" section also conflicts with "Simple JSON Writing and Parsing Example" section later on as both effectively claim to be examples of how to use Oj.

krainboltgreene:
Then why does it crash when I set it to use object mode? :P

ohler55:
I have no idea why it crashes without having more information. What code are you using? Are you dumping or loading? 

As for the default options, try this and note no change in the ident value:
```ruby
irb(main):001:0> Oj.default_options
=> {:indent=>0, :second_precision=>9, :circular=>false, :class_cache=>true, :auto_define=>false, :symbol_keys=>false, :bigdecimal_as_decimal=>true, :use_to_json=>true, :nilnil=>false, :allow_gc=>true, :quirks_mode=>true, :float_precision=>15, :mode=>:object, :escape_mode=>:json, :time_format=>:unix_zone, :bigdecimal_load=>:auto, :create_id=>"json_class"}
irb(main):002:0> Oj.default_options.merge!({:indent => 3})
=> {:indent=>3, :second_precision=>9, :circular=>false, :class_cache=>true, :auto_define=>false, :symbol_keys=>false, :bigdecimal_as_decimal=>true, :use_to_json=>true, :nilnil=>false, :allow_gc=>true, :quirks_mode=>true, :float_precision=>15, :mode=>:object, :escape_mode=>:json, :time_format=>:unix_zone, :bigdecimal_load=>:auto, :create_id=>"json_class"}
irb(main):003:0> Oj.default_options
=> {:indent=>0, :second_precision=>9, :circular=>false, :class_cache=>true, :auto_define=>false, :symbol_keys=>false, :bigdecimal_as_decimal=>true, :use_to_json=>true, :nilnil=>false, :allow_gc=>true, :quirks_mode=>true, :float_precision=>15, :mode=>:object, :escape_mode=>:json, :time_format=>:unix_zone, :bigdecimal_load=>:auto, :create_id=>"json_class"}
irb(main):004:0>
```

krainboltgreene:
Well don't I feel red in the face, looks like we changed it recently:

``` ruby
Oj.default_options = Oj.default_options.merge({
  escape_mode: :ascii
})
```

krainboltgreene:
Check out the rewrite and let me know what you think.

ohler55:
No worries.

I like it better but as long as there is a re-org going on how about moving the Installation section to right under the Using section. After that do into the description. Add a Description heading or something to that effect. Description doesn't feel like exactly the right term.

The on the section for describing options, tighten it up so that it is not code but a set of list items. As code lines trail off to the right and it would look nicer with different formatting. Same information but more nicely formatted is better.

krainboltgreene:
Yes! I tend to do using -> installing as well. Very good point.

ohler55:
Are you still working on changes?

krainboltgreene:
Yep, sorry, work has been busy. I'll do it tonight.

ohler55:
Not really a rush. Just checking.

krainboltgreene:
Give the preview a shot, let me know.

ohler55:
It does look cleaner. I like that. I still want to get the options in something other than code so that the descriptions show up without a horizontal scroll. It really doesn't need to be code and would be easier to read if it wasn't. The rest is good to go though.

ohler55:
Are you willing to make the changes? If not I might take what you have done and then change it immediately.

ohler55:
Merged but will change shortly.

PR #271, opened 2015-11-07
constify RStruct access

nobu:
`RSTRUCT_PTR()` makes the struct WB-unprotected, and causes GC penalty.


ohler55:
Can you explain in more detail why RSTRUCT_CONSt_PTR is better?

Tests are failing with that change. I think you need a cast.

nobu:
`CONSt` is a just typo.
WB-unprotected object requires full-marking of its children, even if they are aged.
It increases marking time of GC.

ohler55:
ありがと

PR #272, opened 2015-11-16
fix parsing big numbers with trailing zeros

comboy:
Hi,

I stumbled upon this:

    >> Oj.load '[10000000000000000000]'
    => [-8446744073709551616]

And I think I managed to find and fix it (substracting <tt>zero_cnt</tt> seems to only make sense for the decimal part, and <tt>LONG_MAX</tt> check doesn't catch it because one iteration it's below and the next one it's already overflowed).

Couldn't calculating the actual number <tt>ni.i</tt> be removed completely here?

Also, there is the same code part in sparse.c which I did not change. I'm assuming it's for strict parsing? But when playing a bit it seemed that parse.c was also used for strict. So since I didn't know how to test it, I left it alone even though probably it needs the same fix.

PS. Awesome work with the lib, I've had an amazing performance improvement after switching to it some time ago.

### DIFF
--- a/ext/oj/parse.c
+++ b/ext/oj/parse.c
@@ -441,7 +441,7 @@ read_num(ParseInfo pi) {
 		}
 		// TBD move size check here
 		ni.i = ni.i * 10 + d;
-		if (LONG_MAX <= ni.i || DEC_MAX < ni.dec_cnt - zero_cnt) {
+		if (LONG_MAX <= ni.i || DEC_MAX < ni.dec_cnt) {
 		    ni.big = 1;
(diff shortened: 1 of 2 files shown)


ohler55:
Good catch. Looking it over there are a few thing that should be done to get it right. Some of the parts of from old ideas that are not being used. I'll take the PR but make some changes after that.

ni.i is needed. It is where the value of a fixnum is stored and passed to callback layer.

Glad you like the gem.

comboy:
Wow, you're fast. Thanks!

PR #280, opened 2015-12-25
the second parameter of gettimeofday() should be NULL

rhenium:
Oj's native extension doesn't compile on Solaris and Ruby 2.3.0.
This is because as of r52703, `-D_XOPEN_SOURCE=xxx` may be specified on Solaris.


ohler55:
Thanks. I probably should not have that test file in the mix anyway.

PR #283, opened 2016-01-05
RB_GC_GUARD to protect from premature GC

johnsyweb:
Context
-----------

I have been auditing some third-party Gems for memory errors in our app, which I believe are caused by [premature GC in C extensions](https://github.com/ruby/ruby/blob/trunk/doc/extension.rdoc#appendix-e-rb_gc_guard-to-protect-from-premature-gc).

Change
------------

In each of these writer functions, `self` is passed by value but isn't referenced after we take a pointer to it (`sw`), so could be optimised away after that. Allocs (or reallocs) could cause a GC to clean up `self`, invalidating `sw`. I introduce `RB_GC_GUARD(self);` later in the function to prevent this.

### DIFF
--- a/ext/oj/oj.c
+++ b/ext/oj/oj.c
@@ -1102,6 +1102,7 @@ str_writer_push_object(int argc, VALUE *argv, VALUE self) {
 	rb_yield(Qnil);
 	oj_str_writer_pop(sw);
     }
+    RB_GC_GUARD(self);
     return Qnil;
 }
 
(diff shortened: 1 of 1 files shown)


johnsyweb:
Ignore me.

ohler55:
The self VALUE is on the Ruby call stack so will not be GCed. That is true of any argument on the Ruby call stack. If the object were allocated in the C function that would be a potential problem with being GCed early but that is not the case in the changes you made.

ohler55:
Race condition. Commented at the same time. :-)

johnsyweb:
Thank you for reviewing this PR so quickly! :-)

ohler55:
Just happened to be at my desk. Alway good to have people reviewing code. Thanks.

Issue #285, opened 2016-02-08
Oj skips colons in strings


thedrow:
```
2.2.0 :001 > require 'oj'
 => true
2.2.0 :002 > Oj.load('{"k": ":"}')
 => {"k"=>:""}
```
This prevents us from deploying Oj to production.

ohler55:
You are using the default :object mode. Change to :compat and it will return a string.
``` ruby
irb(main):001:0> Oj.load('{"k": ":"}', mode: :compat)
=> {"k"=>":"}
```

ohler55:
Or you can change the default_options.

thedrow:
Why is this the correct behavior? Colons are allowed in JSON strings.

ohler55:
Oj has several modes. One, the default is for serializing Ruby objects. It produces valid JSON but has conventions to allow it to create the original objects. http://ohler.com/oj/Oj.html describes the modes. The format used to encode objects is described on http://www.ohler.com/dev/oj_misc/encoding_format.html. Generally you should not try to write those strings directly. Let Oj do it if thats the way you want to use it.

The compat mode loads as native types only so can not be used for direct and optimized object marshaling.

thedrow:
I still don't understand why colons inside a value of an object's key should be omitted in object mode.
```{"k":":-)"}``` is valid JSON and should be parsed into { :k => ':-)' } at all modes. You are losing information in a non-obvious way.

thedrow:
I have an example of a JSON that omits ':' inside string values in compat mode as well but I can't share it here since it contains actual production data.
Unfortunately when I scrub the sensitive data out of it, the test I wrote passes.
@ohler55 Can I please send it to you by email?

ohler55:
Email would be fine. As for the :object mode, the information is not dropped but if the key does not match the class of object being created it can not be set and if the object class is not set then the rules for converting certain syntax to specific types is followed. The :object is really only for object marshaling with Oj on both ends. Trying to hand craft the JSON for loading is not the intent.

If you can share the :compat example that would be create. The conversion or loss should not occur there unless you are exercising the 'json_class' feature of the Ruby JSON library. If that is the case and it is not what you want try the :strict mode.

thedrow:
But that means that if Oj serializes a value with a colon in it in object mode it will be dropped. It doesn't make any sense as a sane default.
It also doesn't make any sense if you are serializing an object that has a string attribute with a value of ':-)' in it like in my case.
[...]

PR #286, opened 2016-02-08
Added a failing test for loading an object with a colon inside a string value

thedrow:
This demonstrates #285.


ohler55:
The added test does not represent the documented nor the desired behavior.

PR #288, opened 2016-02-09
Compile extension in parallel

thedrow:
Travis workers has two cores. We should use them if possible.


thedrow:
Yup that makes the build much faster. You can merge :)

ohler55:
Thanks for the speedup.

PR #289, opened 2016-02-11
Added a note about :object mode

### LINKED ISSUE #258: rounding a float result a different float with more precisions in to_json
the problem in oj 2.12.12:
[1] pry(main)> a = 0.8136100000000001 
=> 0.81361
[2] pry(main)> a.round(5).to_json
=> "0.8136100000000001"

in version  2.9.3, the to_json method will return correct result "0.81361"

thedrow:
I described why the :object mode doesn't fit for anything other than what was serialized with Oj itself.
Fixes #258.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ puts "Same? #{h == h2}"
 # true
 ```
 
+**Attention**
+By default Oj uses the `:object` mode which is used to marshal and unmarshal Ruby objects.
+Do not attempt to deserialize arbitrary JSON in object mode. This may result in issues like in [#285](https://github.com/ohler55/oj/issues/285).
+Instead use the `:compat` or the `:strict` mode.
+
 ## Installation
 ```
 gem install oj


ohler55:
I know you are not happy with the object mode but a warning at the start is not the place to express your dissatisfaction. I'd be glad to expand on the fact that formats are different in each mode further down in the options section. I would prefer to write it myself though and mention that all modes have compatibility issues depending on the use.

Alternatively I could add a different page that goes into the format differences of each instead of just the current one page for the object mode format.

thedrow:
It's not that I'm not happy. I just don't want others to have the same problem as I did.
It's not personal. I hope you didn't take offense.
It's not obvious from the current documentation that object mode doesn't fit for general usage but just for marshaling. If that's going to be the default I'm fine with it as long as people will immediately check the options available to them in order to use Oj for general JSON deserialization.

ohler55:
I'll put something in place today. Let me know if it is clear enough after that. If not we can make another pass at it.

ohler55:
I updated the Readme and the encoding page. Please take a look.

thedrow:
It is clear but it's still not obvious that you shouldn't be using the object mode for anything else but marshalling. I think that this pull request will prevent others from making the same mistake that I did.

ohler55:
The problem with putting the warning "front and center" is that over the years many people have run into problems that they thought were the most important. If Every one of those misunderstanding had a warning place at the top of the page there would be tens if not a hundred warnings about what not to do wrong. It would make finding the one problem you care about even harder to find mixed in with all the other critical misunderstandings. You are the first person who has brought this issue up in the years that Oj has been available. In my mind that does not put it at the top of the list.

I am all for making it obvious that people should consider what mode they are using but not before even getting to the part about installing Oj. If it would help I can make the link in the options section in bold. That would draw the eye to it but in the end the options section is really the right place for it.

quentindemetz:
After struggling around this issue for a while I think the best behaviour would be to use something else than :object mode as default. :strict would make more sense IMHO.

ohler55:
If it was the first release I'd probably agree. Unfortunately the way Oj started out was as a replacement for YAML marshaling. At this point changing the default would break a lot of people code. I can't in good conscience make a change to the default for one person that would cause that many problems. I'm not saying I disagree that strict would be a better default but only that from where Oj stands today making the change is not the right choice.

quentindemetz:
Totally understood!

ohler55:
How about this instead.


 By default Oj uses the `:object` mode which is used to marshal and unmarshal Ruby objects.
 Deserialize arbitrary JSON in object mode make lead to unexpected results. `:compat` mode is a better choice for rails and `:strict` mode is a better choice to general JSON parsing. See the options section below for details.

thedrow:
Looks good.

quentindemetz:
+1

ohler55:
Would you like to update this PR or just drop it and I'll make the changes?

thedrow:
Just drop the PR and make the change.

PR #299, opened 2016-04-30
Render section title by bold font

dtan4:
Modify README section title to render properly.


ohler55:
Thanks. The renderer I normally use displayed correctly. I'll have to be more careful in the future.

PR #304, opened 2016-06-20
Support to Integer unification for Ruby 2.4. 

hsbt:
Ruby 2.4 will unify Bignum and Fixnum to Integer. ref. https://bugs.ruby-lang.org/issues/12005

I couldn't build oj-2.16.0 on ruby 2.4 with folloing error.

```
compiling ../../../../ext/oj/fast.c
../../../../ext/oj/fast.c:1284:24: error: use of undeclared identifier 'rb_cFixnum'
        case T_FIXNUM:  type = rb_cFixnum;      break;
                               ^
1 error generated.
gmake: *** [Makefile:241: fast.o] Error 1
```

I fixed this to use `RUBY_INTEGER_UNIFICATION` provided from Ruby 2.4.


ohler55:
Thank you. I haven't found the time to try Ruby 2.4.

PR #323, opened 2016-12-14
Add CHANGELOG.md

chulkilee:


chulkilee:
Here is a script I used to generate it from web page

See https://gist.github.com/chulkilee/6361d26130fcf76fbda02c872238fade

ohler55:
Nice but can you change it to use Ox instead?

chulkilee:
Do you want to add the conversion script to the repo? Otherwise, I'd let it as is since it just works :)

ohler55:
It would be great if you can add it. That way I can make it part of my release process. I hate to use nogogiri though since Ox id s sibling of Oj.

chulkilee:
Hm.. I believe CHANGELOG.md file in the repo must be source of [the HTML page](http://www.ohler.com/dev/oj_misc/release_notes.html), not vice versa.

I added a commit which makes yard doc include CHANGELOG.md - you may make the html page to link to CHANGELOG.md on your choice (github, rubydoc.info, or ohler.com).

ohler55:
Let me look at integrating the release notes page on ohler.com with the change log. Will you update and add the script to this PR or is that something you prefer me to do?

chulkilee:
You should convert this CHANGELOG.md to html, not vice versa, since this CHANGELOG.md should be source of the html. So the script is no longer be needed.

PR #341, opened 2017-03-15
Do not use a before_sep in pretty_generate

ianks:
The JSON module does not use any before separator when pretty generating JSON. This can be shown with this command:

```
$ ruby -rjson -e "puts JSON.pretty_generate(test: true, hello: 'world')"
{
  "test": true,
  "hello": "world"
}
```

This patch ensures that no `before_sep` is used when calling `pretty_generate`. Although the change is trivial, I think it is important for users who wish to transition to `Oj`. They will not have to deal with breaking any assertions they may have about the output of JSON in their test suites. I noticed this while working on https://github.com/ianks/yagni_json_encoder, which currently passes the entire Rails test suite minus this spec.


ohler55:
I guess when it was first written the default was different. Thanks for keeping it up to date.

PR #342, opened 2017-03-15
Add option to raise Exception on empty strings

ianks:
Many JSON parsers (including Ruby's JSON module and Javascripts `JSON.parse`),
raise exceptions when only whitespace is given as an input string. This patch
adds an optional configuration parameter to raise an exception when an empty
string is given as an input.

This change also updates the Mimic configuration to use this by default, which
corresponds to the behavior of `JSON.parse`. This can be illustrated by running:

```
$ ruby -rjson -e "JSON.parse('   ')"
/home/ianks/.gem/ruby/2.3.1/gems/json-2.0.3/lib/json/common.rb:156:in `parse': 743: unexpected token at '' (JSON::ParserError)
	from /home/ianks/.gem/ruby/2.3.1/gems/json-2.0.3/lib/json/common.rb:156:in `parse'
	from -e:1:in `<main>'
```

Oj now raises a similar exception:

```
$ ruby -roj -e "Oj.mimic_JSON; JSON.parse('   ')"
-e:1:in `parse': unexpected character at line 1, column 3 [parse.c:606] (JSON::ParserError)
	from -e:1:in `<main>'
```

This was the only difference I witnessed when attempting to pass all of the `ActiveSupport::JSON.decode` specs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,10 @@ Oj.default_options = {:mode => :compat }
  * `:nilnil` [Boolean] if true a nil input to load will return nil and
    not raise an Exception
    
+ * `:empty_string` [Boolean] if true an empty input will not raise an
+   Exception, default is true (allow). When Oj.mimic_JSON is used,
+   default is false (raise exception when empty string is encountered)
+
  * `:allow_gc` [Boolean] allow or prohibit GC during parsing, default is
    true (allow).
    
--- a/ext/oj/oj.c
+++ b/ext/oj/oj.c
@@ -130,6 +130,7 @@ static VALUE	mode_sym;
 static VALUE	nan_sym;
 static VALUE	newline_sym;
 static VALUE	nilnil_sym;
+static VALUE	empty_string_sym;
 static VALUE	null_sym;
 static VALUE	object_sym;
 static VALUE	omit_nil_sym;
@@ -184,6 +185,7 @@ struct _Options	oj_default_options = {
     No,		// to_json
     No,		// as_json
     No,		// nilnil
+    Yes,	// empty_string
     Yes,	// allow_gc
     Yes,	// quirks_mode
     No,		// allow_invalid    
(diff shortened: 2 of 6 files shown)


ohler55:
You have covered the changes in the code well and kept with my formatting. Thank you for that. A question and a request first though.

What is the return value when given an empty string to parse? I would assume nil.

The request is to add tests in test_various for an empty string that is allowed and not allowed. I saw the test in isolate to cover the rails use but would like to have tests for non-rails as well.

ianks:
> What is the return value when given an empty string to parse? I would assume nil.

Are you referring to the current implementation? Yes it is `nil`.

> The request is to add tests in test_various for an empty string that is allowed and not allowed. I saw the test in isolate to cover the rails use but would like to have tests for non-rails as well.

Absolutely. Very good feedback. I was feeling a little iffy about the test cases myself. I will hop on this and add them to the PR.

ianks:
@ohler55 I added a couple more tests. In the test, I am dynamically generating what is considered to be a whitespace string 10 different times, and asserting on it. This works as a way to fuzz out any hidden bugs. If you don't like this method, I can happily update it to use static strings 😄

ohler55:
I'll take a look in the morning and merge. Thanks.

ohler55:
Nicely done. Thank you.

ianks:
Thanks @ohler55 👏 

Which version do you plan to ship this in?

ohler55:
I am working on a set or rails compatibility updates right now. It will go in with those changes. Maybe by the end of the week if all goes well. Might get pushed to next week though. If that is a problem I could do an minor release if necessary.

ianks:
That should be fine! Just wanted to know so I can keep track of it. Thanks again 😀

ohler55:
I release this evening. I was going to wait for the next release but it is turning out to be a big one so it will be 3.0.

PR #343, opened 2017-03-16
Different gemspecs for different rails

stereobooster:
If you want to test compatibility with different rails. The strange thing is that non off rails versions supports ruby 1.8.7, so if you want to continue to support 1.8.7 we will need either switch to earlier version of rails or run tests without rails.

To run tests locally use `wwtd` command


ohler55:
Thanks. That is certainly rails friendly.

PR #344, opened 2017-03-17
Compatibility tests

stereobooster:
Tell me what do you think. I commented out some examples, till they will be fixed in `rails-compat` branch


ohler55:
I noticed one file was in isolated but the others were not. All use active classes so I think it makes more sense to put them all in isolated. Agree?

stereobooster:
Opened PR for rails-combat branch

PR #345, opened 2017-03-17
Updated travis matrix

stereobooster:
I dropped support for ruby 1.8.7 and 1.9.3, because they are not officially supported, and tests were failing. Same for Rails 3


.travis.yml
```diff
@@ -10,35 +10,20 @@ cache:
 
 language: ruby
 rvm:
-  - 1.8.7
-  - 1.9.3
```
stereobooster: We can skip all rails tests if rails are not in Gemfile, like

```ruby
begin
  require "rails/all"
rescue
  puts "skipping rails tests"
  exit
end
```

Plus crate `gemfiles/no_rails.gemfile` and target non-rails ruby versions. Though I want to notice, that ruby 1.9.3 is not supported officially and its use discouraged.

ohler55: Yeah 1.9.3 is not supported but still in use so at least I prefer not to break it if I don't have to.

stereobooster: screen shot 2017-03-19 at 22 21 06

> I prefer not to break it if I don't have to.

From my POV it is broken. I'm not insisting here. Just thinking out loud.

ohler55: It is broken due to running rails tests (I think, I have checked to make sure). Ideally I'd like to have 1.9.3 non-rails tests run only. I'll deal with it after the refactor though unless you feel like taking care of it.

I am fine getting rid of 1.8.7. As it is I am removing the support in the code during the refactoring.

PR #347, opened 2017-03-17
Improve Readme

stereobooster:
Trying to sort options. Plus corrected formatting a bit. It may happen that I do not understand which options fall into which category. Please check.

Want to sort options, to count number of testcases.


ohler55:
Looks good and is very reasonable.

PR #348, opened 2017-03-18
Fix travis

stereobooster:
Just removed lock files


ohler55:
Ah, removed the lock files. Makes sense. Thanks.

PR #349, opened 2017-03-18
More data for tests

stereobooster:
Maybe will find more testcases. Scanning through rails tests

- [encoding_test.rb](https://github.com/rails/rails/blob/e8ba0c0f21e2660b90f872fa4595156ca6190c77/activesupport/test/json/encoding_test.rb)
- [encoding_test_cases.rb](https://github.com/rails/rails/blob/e8ba0c0f21e2660b90f872fa4595156ca6190c77/activesupport/test/json/encoding_test_cases.rb)
- [json_encoding_test.rb](https://github.com/ruby/ruby/blob/trunk/test/json/json_encoding_test.rb)
- [json_generator_test.rb](https://github.com/ruby/ruby/blob/trunk/test/json/json_generator_test.rb)

### DIFF
--- a/test/isolated_compatibility/data_compatibility_json.rb
+++ b/test/isolated_compatibility/data_compatibility_json.rb
@@ -17,6 +17,18 @@ Struct.new('Customer', :name, :address)
 fork { exit 99 }
 Process.wait
 
+class JsonRenderable
+  def as_json(options = {})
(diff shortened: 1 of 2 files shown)


ohler55:
Thanks, keep the test case coming. :-)

PR #350, opened 2017-03-19
Can not push onto an Object without a key

cschlack:
```ruby
w = Oj::StringWriter.new
w.push_object()
w.push_value(nil, 'nothing')
```
=> `Can not push onto an Object without a key. (StandardError)`


ohler55:
Thanks. With tests too, great.

PR #356, opened 2017-03-24
add more tests

stereobooster:
- [encoding_test.rb](https://github.com/rails/rails/blob/e8ba0c0f21e2660b90f872fa4595156ca6190c77/activesupport/test/json/encoding_test.rb)
- [encoding_test_cases.rb](https://github.com/rails/rails/blob/e8ba0c0f21e2660b90f872fa4595156ca6190c77/activesupport/test/json/encoding_test_cases.rb)
- [json_encoding_test.rb](https://github.com/ruby/ruby/blob/trunk/test/json/json_encoding_test.rb)
- [json_generator_test.rb](https://github.com/ruby/ruby/blob/trunk/test/json/json_generator_test.rb)


ohler55:
Thanks for continuing with tests. I think the upcoming 3.0 release will clear up a lot of issues with compatibility.

stereobooster:
Those are really strange tests. Take a look after I will finish. I feel now: o_0

ohler55:
The tests are exactly correct. The JSON gem actually uses 3 modes, JSON.generate(), JSON.dump(), and obj.to_json(). Sadly they differ in behavior. So JSON.generate() generally seems to follow obj.to_json() while dump is a bit different. I ran into this with Infinity.

ohler55:
I am also trying to not use the the Oj options except where they match the JSON options. Further, the options passed to Oj.to_json() and Oj.generate() should be the same as the JSON gem, not the Oj options. I've got to get the mapping for functions set up correctly. Tonight's task along with cleaning up object mode.

stereobooster:
@ohler55 I have crazy idea: what about porting JSON tests?

ohler55:
I like that idea. With mimic_JSON they should all pass at least that could be the goal.

stereobooster:
Kind of ported. Probably some more work required. Take a look, what you want to change @ohler55

ohler55:
Looks reasonable to me. Some work still needed of course but seems like the right direction.

stereobooster:
- I checked that all json_gem tests pass for native implementation.
- I omitted all functionality which is not `to_json`, 'dump' or `load`
- I marked tests as pending, which cause segmentation fault or similar

ohler55:
This might take a while to get them all fixed.

ohler55:
I fixed a few tests in the additions tests. Ran into an issue I'm not sure how to deal with yet. You required the 'json/add/xxx' paths that monkey path the already monkey patched core classes. Oj can offer an alternative to those with something like 'oj/json/add/xxx' but I haven't figured out how to allow core classes to be monkey patched and to tell the difference when JSON() or generate() is used and still get the more optimized C encoding instead. Especially tricky when the behavior has to be different depending on if to_json is called or generate, dump, or JSON.

stereobooster:
> required the 'json/add/xxx'

That is how original tests done. Try comment out those lines and run against native implementation to see if they work.

stereobooster:
I omitted tests for `JSON`. Who use that?

If you see some inconsistency in function calls, I think it would make sense to open bugreports in Ruby tracker (JSON gem part of ruby).

The same way I was thinking about opening report for `[NaN]`, that is not valid JSON. Why it was implemented that way 🙄

ohler55:
It would be nice it the tests passed. There might be a way. I'll be out the rest of the weekend but will give it some thought. If I don't come up with something we can fall back to not using the adds but I suspect others out there use them and maybe rails does as well.

As for NaN, they never said the json gem was JSON compliant so I don't think they will change anything. It was a decision they made a long time ago and now we are stuck with it.

Nice job on the tests though. Other than not passing in travis they provide great coverage. I would like to get rid of or reverse the env var MIMIC_JSON to something like USE_JSON_GEM or something so the default for the tests is to use mimic_JSON and not the son gem.

PR #357, opened 2017-03-30
Yield next offset when parsing multiple objects

krallin:
Pass the next offset as a second block argument when a block is passed
to `Oj.load`.

This allows an end-user to know what had been successfully parsed if an
error occurs after some objects have been parsed successfully, which is
useful if the stream being parsed may be incomplete (i.e. it's being
read from somewhere else that doesn't buffer on a per-object basis).

---

A couple notes though...:

- First, Oj ignores the offset when a `StringIO` is provided, so this API isn't entirely trivial to use: you  need to either pass in a *new* StringIO that starts at the right offset, or `read` the remaining contents of the StringIO. If Oj looked at the `StringIO`'s offset, then a user could just `seek` to the last offset that was successfully parsed, then resume parsing, which would probably be easier to use. I suspect this would be a breaking change though 😢, so let me know what you think!
- I'm not sure how to enter the branch that calls `rb_proc_call_with_block`, so the changes I've made there are completely untested. I'd be happy to actually add test coverage for those if you could let me know in which case this gets used 😄 

Thanks for your time, and thanks for Oj!

### DIFF
--- a/ext/oj/parse.c
+++ b/ext/oj/parse.c
@@ -676,14 +676,16 @@ oj_parse2(ParseInfo pi) {
 	}
 	if (stack_empty(&pi->stack)) {
 	    if (Qundef != pi->proc) {
+		long pos = pi->cur - pi->json;
 		if (Qnil == pi->proc) {
-		    rb_yield(stack_head_val(&pi->stack));
+		    rb_yield_values(2, stack_head_val(&pi->stack), INT2NUM(pos));
 		} else {
 #if HAS_PROC_WITH_BLOCK
-		    VALUE	args[1];
+		    VALUE	args[2];
 
-		    *args = stack_head_val(&pi->stack);
-		    rb_proc_call_with_block(pi->proc, 1, args, Qnil);
+		    args[0] = stack_head_val(&pi->stack);
+		    args[1] = INT2NUM(pos);
+		    rb_proc_call_with_block(pi->proc, 2, args, Qnil);
 #else
 		    rb_raise(rb_eNotImpError,
 			     "Calling a Proc with a block not supported in this version. Use func() {|x| } syntax instead.");
--- a/ext/oj/sparse.c
+++ b/ext/oj/sparse.c
@@ -719,14 +719,16 @@ oj_sparse2(ParseInfo pi) {
 	}
 	if (stack_empty(&pi->stack)) {
 	    if (Qundef != pi->proc) {
+		long pos = pi->rd.tail - pi->rd.head;
 		if (Qnil == pi->proc) {
-		    rb_yield(stack_head_val(&pi->stack));
+		    rb_yield_values(2, stack_head_val(&pi->stack), INT2NUM(pos));
 		} else {
 #if HAS_PROC_WITH_BLOCK
-		    VALUE	args[1];
+		    VALUE	args[2];
 
-		    *args = stack_head_val(&pi->stack);
-		    rb_proc_call_with_block(pi->proc, 1, args, Qnil);
+		    args[0] = stack_head_val(&pi->stack);
+		    args[1] = INT2NUM(pos);
+		    rb_proc_call_with_block(pi->proc, 2, args, Qnil);
 #else
 		    oj_set_error_at(pi, rb_eNotImpError, __FILE__, __LINE__,
 				    "Calling a Proc with a block not supported in this version. Use func() {|x| } syntax instead.");
--- a/test/test_strict.rb
+++ b/test/test_strict.rb
@@ -238,9 +238,61 @@ class StrictJuice < Minitest::Test
   def test_double
     json = %{{ "x": 1}{ "y": 2}}
     results = []
-    Oj.load(json, :mode => :strict) { |x| results << x }
+    Oj.load(json, :mode => :strict) { |x, i| results << [x, i] }
 
-    assert_equal([{ 'x' => 1 }, { 'y' => 2 }], results)
+    assert_equal([[{ 'x' => 1 }, 9], [{ 'y' => 2 }, 18]], results)
+  end
+
+  def test_resume
+    json = %{{ "x": 1}{ "y": 2}}
+
+    objects = []
+    read_offset = 0
+
+    assert_raises(Oj::ParseError) do
+      Oj.load(json[0...13], :mode => :strict) do |obj, offset|
+        objects << obj
+        read_offset = offset
+      end
+    end
+
+    assert_equal([{ 'x' => 1 }], objects)
+
+    Oj.load(json[read_offset..-1], :mode => :strict) do |obj|
+      objects << obj
+    end
+
+    assert_equal([{ 'x' => 1 }, { 'y' => 2 }], objects)
+  end
+
+  def test_resume_io
+    json = %{{ "x": 1}{ "y": 2}}
+
+    objects = []
+    read_offset = 0
+
+    io = StringIO.new
+    io.write(json[0...13])
+    io.seek(0)
+
+    assert_raises(Oj::ParseError) do
+      Oj.load(io, :mode => :strict) do |obj, offset|
+        objects << obj
+        read_offset = offset
+      end
+    end
+
+    assert_equal([{ 'x' => 1 }], objects)
+    assert_equal(9, read_offset)
+
+    io.seek(0, IO::SEEK_END)
+    io.write(json[13..-1])
+    io.seek(read_offset)
+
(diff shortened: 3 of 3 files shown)


ohler55:
With the descriptions you provided I can see how knowing the position in the file, it a file, can help recover. I prefer not to make breaking changes such as what you have done but that doesn't mean there is not a place for the feature in Oj. A load_with_offset or simply Oj.offset_load could be added.

What do you think about including the offset in the exception raised or do you think it is a good idea to know where in a file the JSON element was read from?

Finally, OJ is going through some significant code refactoring on the rails-compat branch right now. I've made changes and will continue to do so for before done on the dump and parse files. Lets discuss what you want here and then it can go into what will become Oj 3.0.

krallin:
Hey there, thanks for your attention!

>  I prefer not to make breaking changes such as what you have done but that doesn't mean there is not a place for the feature in Oj.

I fully agree on avoiding breaking changes.

Just to make sure we're on the same page, though: I haven't actually made any changes to how Oj handles a StringIO input whose current position isn't 0 (which would be a major breaking change).

That being said, adding a new block argument could indeed break some apps if they're using a Lambda rather than a Proc (i.e. checking arity) and therefore fail when receiving 2 block arguments instead of one. Is that an unacceptable breaking change from your perspective?

>  A load_with_offset or simply Oj.offset_load could be added.

How do you envision the API for this? should the offset be passed as an input, or should that function read the offset from a StringIO if it's passed in?

Thanks!

> What do you think about including the offset in the exception raised or do you think it is a good idea to know where in a file the JSON element was read from?

I actually tried doing that initially! 😄 .

The problem there was that it seemed much harder to pull off (but maybe that's just my unfamiliarity with the Oj codebase!). Here's an example:

```
{"foo": "bar"}{"qux": 
```

Here, the parsing will fail once we reach the final character, so if the exception naïvely reports where the failure happened, then that information isn't particularly useful, and we can't use it to resume parsing:

```
> Oj.load('{"foo": "bar"}{"qux"') { }
Oj::ParseError: Hash/Object not terminated at line 1, column 20 [parse.c:908]
```

So, instead, we need to know what the *last valid position* was so that we can resume from it (which is the position immediately following the last valid object).

Here, that's the position of the second `{`.

So, in the end, it seemed easier to me to send that information out to the user when we have it (i.e. when we just completed parsing an object), but I'm totally fine with another approach if it can be pulled off :)

However, if the exception could report the last valid character, then that's perfect 👌 

> Finally, OJ is going through some significant code refactoring on the rails-compat branch right now. I've made changes and will continue to do so for before done on the dump and parse files. Lets discuss what you want here and then it can go into what will become Oj 3.0.

Oh, interesting. How do you want to go about this? I'm happy to share the code I'm currently using that leverages the changes in this PR as a starting point for the conversation — would that help?

Thanks!

ohler55:
Using what you have as a starting point would be helpful. Although at this point putting the changes into the rails-compat branch would be where I would make the changes. I'm pretty sure I could keep the position of the first character in a JSON element for exceptions if that is the better approach. That approach is less flexible though so an additional parameter as you have proposed might be better.

krallin:
Thanks! Here's how I'm using the API exposed by this PR currently:

```
class Reader
  RECYCLE_SIZE = 1024**2 # 1 MB

  class ReadError < StandardError
    def initialize(payload, offset = nil)
      m = "Invalid payload in #{payload}"
      m = "#{m} at offset #{offset}" if offset
      super(m)
    end
  end

  class StateError < StandardError
  end

  def initialize
    @io = StringIO.new
  end

  def continue(chunk, &block)
    # This adds the new chunk at the end of the stream, seeks back to
    # where we were before, and picks up processing from there.
    pos = @io.tell

    # NOTE: A short write isn't ever going to happen here due to how
    # StringIO#write is (currently) implemented, so if one does happen,
    # we don't even try to handle it.
    @io.seek(0, IO::SEEK_END)
    fail 'short write' if @io.write(chunk) != chunk.size

    @io.seek(pos)

    try_read(&block)
  end

  def final
    # We ingest things as we go, so if there's any payload left here,
    # that means we failed to process it earlier, which means it should
    # be invalid JSON. If we somehow manage to get valid JSON here, that
    # means something went very wrong earlier.
    payload = @io.read
    fail StateError, 'Leftover payload' if Oj.strict_load(payload)
  rescue Oj::ParseError
    raise ReadError.new(payload)
  ensure
    @io = nil # won't be able to use this again by mistake
  end

  private

  def try_read
    pos = @io.tell
    next_offset = 0

    # Can't use @io directly since Oj won't look at the position and read it from the beginning.
    payload = @io.read

    Oj.strict_load(payload) do |message, offset|
      fail ReadError.new(payload, next_offset) unless message.is_a?(Hash)
      yield message
      next_offset = offset
    end
  rescue Oj::ParseError
    # Not everything could be read. Presumably we have an incomplete object
    @io.seek(pos + next_offset)
    @io = StringIO.new if @io.eof? && @io.size > RECYCLE_SIZE
  end
```

The way you'd use this is:

```
r = Reader.new
r.continue(get_some_json_1) { |m| do_something_with_the_message(m) }
r.continue(get_some_json_2) { |m| do_something_with_the_message(m) }
r.continue(get_some_json_3) { |m| do_something_with_the_message(m) }
r.final
```

(of course, in our case,  we don't have all the JSON initially, otherwise this whole thing isn't useful)

krallin:
Thanks! Here's how I'm using the API exposed by this PR currently:

```ruby
class Reader
  RECYCLE_SIZE = 1024**2 # 1 MB

  class ReadError < StandardError
    def initialize(payload, offset = nil)
      m = "Invalid payload in #{payload}"
      m = "#{m} at offset #{offset}" if offset
      super(m)
    end
  end

  class StateError < StandardError
  end

  def initialize
    @io = StringIO.new
  end

  def continue(chunk, &block)
    # This adds the new chunk at the end of the stream, seeks back to
    # where we were before, and picks up processing from there.
    pos = @io.tell

    # NOTE: A short write isn't ever going to happen here due to how
    # StringIO#write is (currently) implemented, so if one does happen,
    # we don't even try to handle it.
    @io.seek(0, IO::SEEK_END)
    fail 'short write' if @io.write(chunk) != chunk.size

    @io.seek(pos)

    try_read(&block)
  end

  def final
    # We ingest things as we go, so if there's any payload left here,
    # that means we failed to process it earlier, which means it should
    # be invalid JSON. If we somehow manage to get valid JSON here, that
    # means something went very wrong earlier.
    payload = @io.read
    fail StateError, 'Leftover payload' if Oj.strict_load(payload)
  rescue Oj::ParseError
    raise ReadError.new(payload)
  ensure
    @io = nil # won't be able to use this again by mistake
  end

  private

  def try_read
    pos = @io.tell
    next_offset = 0

    # Can't use @io directly since Oj won't look at the position and read it from the beginning.
    payload = @io.read

    Oj.strict_load(payload) do |message, offset|
      fail ReadError.new(payload, next_offset) unless message.is_a?(Hash)
      yield message
      next_offset = offset
    end
  rescue Oj::ParseError
    # Not everything could be read. Presumably we have an incomplete object
  ensure
    @io.seek(pos + next_offset)
    @io = StringIO.new if @io.eof? && @io.size > RECYCLE_SIZE
  end
```

The way you'd use this is:

```
r = Reader.new
r.continue(get_some_json_1) { |m| do_something_with_the_message(m) }
r.continue(get_some_json_2) { |m| do_something_with_the_message(m) }
r.continue(get_some_json_3) { |m| do_something_with_the_message(m) }
r.final
```

(of course, in our case,  we don't have all the JSON initially, otherwise this whole thing isn't useful)

ohler55:
It sounds like the arguments to the block should be JSON, offset of start, and length. That way you know where the broken segment started.

krallin:
Sure; that'd be equivalent I think: right now, I'm yielding the object and _the end of the segment for that object_, so to know where the broken segment started, I just need to remember where the last valid segment ended 😄 

That being said, I'm totally in favor of exposing more information as you're suggesting — that'd be just as usable for me.

ohler55:
Sounds like a plan then. I am hoping to get 3.0 out in a week, maybe 2. I'll have you change in it even if I can't really take this PR due to the current code restructuring. I'll ping you when I have something you can try.

krallin:
Sounds great! Looking forward to it! 😄

ohler55:
I added the feature in the rails-compat branch it yout want to try it out. If it looks okay then we can close this PR.

krallin:
This is perfect; thanks @ohler55!

PR #358, opened 2017-04-06
ActiveSupport tests

stereobooster:
tell me what else to change

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -57,7 +57,7 @@ task :test_all => [:clean, :compile] do
       exitcode = 1 unless status
     end
   end
-  
+
   if false
   # run JSON tests for Oj.mimic_JSON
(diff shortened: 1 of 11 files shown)


ohler55:
I can tell already that this will keep me busy for a while getting all the tests to pass.

stereobooster:
Well I ported test for 4.1 and 5.0. There is also 4.2 which is popular. But those tests run for `ActiveSupport`, not `Oj` implementation.

ianks:
@ohler55 I have all of these tests passing (minus a few to html escaping stuff) at https://github.com/ianks/yagni_json_encoder

ohler55:
Very nice. I'm continuing to work on getting 100% rails compatibility with optimization. The encoding is the sticking point today. Should get that dealt with soon.

PR #361, opened 2017-04-20
Clean up

stereobooster:
WIP

Theoretically we do not need `isolated_compatibility` tests.
Other small fixes.
I want try to bring JSON gem tests for ruby < 2.4 to see how exactly it is incompatible.


ohler55:
All make sense, thanks for the cleanup.

PR #362, opened 2017-04-20
Try tests for other ruby versions

stereobooster:
Just checking. Do not merge yet

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -28,19 +28,14 @@ task :test_all => [:clean, :compile] do
   end
   exitcode = 1 unless status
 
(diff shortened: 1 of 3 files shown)


ohler55:
I found that ActiveSupport 5 did not pass the ActiveSupport 4 tests so unless Oj has a switch for 4 or 5 it can not pass both sets of tests. I'd rather get the 3.0 release out now and add switches for older versions after that. Do you think 100% ActiveSupport 4 is required before the release?

stereobooster:
> Do you think 100% ActiveSupport 4 is required before the release?

No. Not at all. Just poking around

PR #363, opened 2017-04-21
Fix memory usage for strings

snoopie:
Hi!

Could you please have a look at this fix rather soon?

It fixed a nasty bug where memory usage for string conversions was 49ish times bigger than it should've been. It's because ```*_friendly_size``` methods were modifying input parameter value of ```len``` and final return value calculation always subtracted zero which lead to a huge over-allocation for ```Out``` string.

Here is the comparison of the memory usage before and after fix for ```{body: body}``` hash object where body value is a ```2097152``` characters long ```String```:

Before: https://www.dropbox.com/s/i9p07edj2m61ex6/Memory_usage_before.png?dl=0
After: https://www.dropbox.com/s/mpggchzwldsjma5/Memory_usage_after.png?dl=0

Values above are in kilobytes.

I also took the liberty of removing trailing whitespaces (in a separate commit), let me know if you don't like it and I should remove it.

Thanks!

### DIFF
--- a/ext/oj/dump.c
+++ b/ext/oj/dump.c
@@ -157,30 +157,30 @@ static char	xss_friendly_chars[256] = "\
 33333333333333333333333333333333";
 
 inline static size_t
-newline_friendly_size(const uint8_t *str, size_t len) {
-    size_t	size = 0;
+newline_friendly_size(const uint8_t *str, const size_t len) {
+    size_t	size = 0, counter = len;
 
-    for (; 0 < len; str++, len--) {
+    for (; 0 < counter; str++, counter--) {
 	size += newline_friendly_chars[*str];
     }
     return size - len * (size_t)'0';
 }
 
 inline static size_t
-hibit_friendly_size(const uint8_t *str, size_t len) {
-    size_t	size = 0;
+hibit_friendly_size(const uint8_t *str, const size_t len) {
+    size_t	size = 0, counter = len;
 
-    for (; 0 < len; str++, len--) {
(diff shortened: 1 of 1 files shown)


ohler55:
I'm afraid I can not accept these changes. You are correct that there was a bug in the some of the length calculations but those have been fixed in the rails-compat branch that is due to go out as Oj 3.0 this weekend. Any changes to master at this point would only cause a merge conflict as there have been major code reorganizations.

ohler55:
Oh, on the trailing whitespace. I assume you meant in the source files. Don't bother with that. It is an artifact of the editor I use (emacs) so they will just return as soon as I edit the file again. They are harmless and have no impact on the compiled gem.

snoopie:
@ohler55 How about merging it w/o trailing space removal(like it is now)? It is pretty small and pure change (fix of a bug which should not affect functionality). If we have to wait for 3.0 and then upgrade it would be too risky because 3.0 has many other changes.

We could of course use our own fork, but I would like to avoid it.

snoopie:
@ohler55 I can even align this change with ```rails-compat``` branch (variable names etc)

ohler55:
As I mentioned earlier, the bug is already fixed in the rail-combat branch. Take a look if you like. The fix was similar you yours. The only difference being variable names. That and no const len argument since len is pass by value anyway so const does nothing.

snoopie:
okay

ohler55:
I appreciate the effort. Just bad timing. Funny no one had caught the problem before you. I ran into it by accident when doing the reorg.

PR #366, opened 2017-04-25
Remove calls to `dump_cstr` and `obj_dump_comp_val`

DylanJ:
I was not able to `gem install oj` inside of a docker container because of the error:

```
Error relocating /usr/local/bundle/gems/oj-3.0.0/lib/oj/oj.so: dump_cstr: symbol not found - /usr/local/bundle/gems/oj-3.0.0/lib/oj/oj.so
```

After some digging i noticed the functions `dump_cstr` and `obj_dump_comp_val` were externed but not defined anywhere. I guess these were optimized away at link time?

This is just a quick fix and honestly I'm suprised it fixed my problem. Maybe I broke something but the tests seem to work.


ohler55:
Thanks. StringWriter still needs something but I'll fix that after the merge.

PR #377, opened 2017-05-06
Appveyor badge

stereobooster:


ohler55:
I think it is best to get the appveyor worked out before putting the badge up. Given the segfaults are occurring on a simple ruby call I'm not convinced the environment is set up yet. Lets keep this in reserve until the appveyor stuff is dealt with.

ashmaroli:
Hi, I tried running the tests on my local Windows machine (Windows 7, Ruby 2.3.1) and the tests segfaults occur at places that involve `streams` (`:load_file`, etc), `:time_format => :unix_zone`, and `fork()`

ohler55:
Have you tested with any other Windows versions?

ashmaroli:
sorry, no.. Just this one..
But Oj works great on it however..

ohler55:
Maybe the trick is to exclude tests on Windows and note the limitations.

ashmaroli:
> Maybe the trick is to exclude tests on Windows

As long as they're not testing a fundamental implementation, then it sounds like a plan. Though, it would require guards to ensure things working now don't break in future.

ohler55:
If you know which tests fail that would help me get a head start on the exclusions.

ashmaroli:
I was able to bypass segfaults by `skip`-ping `test_file.rb#dump_and_load`, `test_object.rb#test_time_early`

ohler55:
Thanks.

ohler55:
Working through the tests. The early time is fixed, sort of. Windows does not support dates before 1970 so that test will never pass on Windows. Same with fork. Anyway, working now so will merge this PR.

ohler55:
Finally got appveyer working so now it makes sense to put up the badge.

PR #380, opened 2017-05-08
Fix typo

frodsan:


frodsan:
I created an initializer in my Rails project like this:

```
# frozen_string_literal: true

# Be sure to restart your server when you modify this file.

Oj.default_options = {
  mode: :rails,
  allow_blank: true
}
```
Do I need to call `Oj.optimize_rails` too?

Thanks.

ohler55:
Thanks for catching that.

@frodsan You should call Oj.optimize_rails or call Oj::Rails.set_encoder and decoder. In Rails the ActiveSupport.json_encoder is used for encoding. By setting that to point at Oj all JSON encoding will use Oj.

You should also take the latest version of Oj as after a major release there are usually a number of bug fixes that come out soon after.

frodsan:
@ohler55 Thanks for response. I'm confused by the Rails integration guide. So, to properly integrate Oj into Rails, I need an initializer like this:

```
# config/initializers/oj.rb
require 'oj'

Oj.default_options = {
  mode: :rails,
  allow_blank: true
}

Oj.optimize_rails
```

Thanks again.

ohler55:
All you need is this.

```ruby
# config/initializers/oj.rb
require 'oj'

Oj.optimize_rails
```

PR #385, opened 2017-05-16
Fix relative links

mortonfox:


ohler55:
Thanks. Missed that since the yardoc output works correctly.

PR #395, opened 2017-05-31
Remove call to suggest image from README

hmac:
Looks like this is now done!


ohler55:
Yes, it is. Thanks for the cleanup.

PR #397, opened 2017-06-05
Fix typo in method name

misterbyrne:
The `Oj::Rails::mimic_JSON` method is being defined with a trailing question mark. The documentation suggests that this isn't intentional:

https://github.com/ohler55/oj/blob/master/pages/Rails.md


ohler55:
You are completely correct. Time for a fresh release.

PR #415, opened 2017-08-07
Replace global variables with instance variables

yui-knk:


ohler55:
This was not that important to me but it does keep the variable separate from other tests.

yui-knk:
Thanks :)

PR #416, opened 2017-08-08
Remove RUBY_VERSION check for ruby 1.8 and 1.9

yui-knk:
Currently we do not support these ruby versions and do not
run tests against them.
So we do not need to check RUBY_VERSION for ruby 1.8 and 1.9.


ohler55:
Long over due. Thank you.

You have been active here recently. I see you are also helping with Rails and Ruby. Are you interested in helping on the WABuR (https://github.com/ohler55/wabur) project I recently started?

PR #418, opened 2017-08-10
Update oj.gemspec to include a minimum ruby version

knappe:
Add minimum ruby version to gemspec, so it can be reported by rubygems.


ohler55:
Thanks

PR #426, opened 2017-08-15
Rename `UNIFY_FIXNUM_AND_BIGNUM` macro

yui-knk:
We use this macro only to check we should cast returned value of
`RSTRUCT_LEN` to `long`. So `RSTRUCT_LEN_RETURNS_INTEGER_OBJECT`
is better.
By the way, we can use `RUBY_INTEGER_UNIFICATION` macro to detect
Integer is unified or not.


ohler55:
Thanks for you help getting Oj up to date with the latest Ruby.

PR #427, opened 2017-08-16
Drop `DATETIME_1_8` macro

yui-knk:
We do not support Ruby 1.8, so we can remove this macro.


ohler55:
Nice cleanup. Thanks.

PR #428, opened 2017-08-18
Avoid "cannot load such file -- test/unit (LoadError)"

yui-knk:


ohler55:
You are being an amazing help. Thanks you.

PR #429, opened 2017-08-23
Use `rb_scan_args` to check args

yui-knk:
If we pass `"11"` to 3rd argument of `rb_scan_args`,
`rb_scan_args` raises `ArgumentError` when `argc` is not
1 nor 2.
This commit changes to raise `ArgumentError` if `argc > 2`.

### DIFF
--- a/ext/oj/mimic_json.c
+++ b/ext/oj/mimic_json.c
@@ -485,9 +485,10 @@ oj_mimic_pretty_generate(int argc, VALUE *argv, VALUE self) {
 static VALUE
 mimic_parse_core(int argc, VALUE *argv, VALUE self, bool bang) {
     struct _ParseInfo	pi;
+    VALUE		ropts;
     VALUE		args[1];
 
-    rb_scan_args(argc, argv, "11", NULL, NULL);
+    rb_scan_args(argc, argv, "11", NULL, &ropts);
     parse_info_init(&pi);
     oj_set_compat_callbacks(&pi);
     // TBD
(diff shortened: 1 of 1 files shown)


yui-knk:
Thanks for your review.

Sure rb_scan_args may be slower than checking the args directly, because rb_scan_args parses fmt, ""11"".
I think  the advantage of rb_scan_args is this function ensures count of arguments is 1 or 2.
For example:

## Before

```
irb(main):001:0> require "oj"
=> false
irb(main):002:0> Oj.mimic_JSON
=> JSON
irb(main):003:0> JSON.parse("{}", {}, 2) rescue p $!
=> {}
```

## After

```
irb(main):001:0> require "oj"
=> false
irb(main):002:0> Oj.mimic_JSON
=> JSON
irb(main):003:0> JSON.parse("{}", {}, 2) rescue p $!
#<ArgumentError: wrong number of arguments (given 3, expected 1..2)>
=> #<ArgumentError: wrong number of arguments (given 3, expected 1..2)>
```

This behavior is compatible with the behavior of json gem.

```
irb(main):001:0> require "json"
=> true
irb(main):002:0> JSON.parse("{}", {}, 2) rescue p $!
#<ArgumentError: wrong number of arguments (given 3, expected 1..2)>
=> #<ArgumentError: wrong number of arguments (given 3, expected 1..2)>
```

ohler55:
If it becomes a performance issue I'll go back to the more direct approach. Thank you.

PR #445, opened 2017-10-27
Odd char

ohler55:


gerrywastaken:
I really like the addition of the extra debug info by default. Cheers for that.

ohler55:
I figured it might come in useful someday again.

PR #458, opened 2018-02-15
gemspec: add metadata URLs, drop rubyforge_project

olleolleolle:
Hi, and thanks for `oj` for all this time.

I looked at the [metadata](http://guides.rubygems.org/specification-reference/#metadata), and added a few links. These would show up in the RubyGems website interface, and also be neat metadata.


ohler55:
It was in need of an update. Thank you.

Issue #459, opened 2018-02-23
Tracer.display_c_call doesn't show OJ


nateberkopec:
```ruby
require 'oj'
Oj.mimic_JSON()
require 'tracer'
Tracer.display_c_call = true
Tracer.on

Math::PI.to_json
Oj.dump(Math::PI)
```

Output is:
```
#0:oj_verify.rb:7::-: Math::PI.to_json
#0:oj_verify.rb:7:Object:}: Math::PI.to_json
#0:oj_verify.rb:7:Float:}: Math::PI.to_json
#0:oj_verify.rb:7:Float:{: Math::PI.to_json
#0:oj_verify.rb:7:Object:{: Math::PI.to_json
#0:oj_verify.rb:8::-: Oj.dump(Math::PI)
#0:oj_verify.rb:8:Oj:}: Oj.dump(Math::PI)
#0:oj_verify.rb:8:Float:}: Oj.dump(Math::PI)
#0:oj_verify.rb:8:Float:{: Oj.dump(Math::PI)
#0:oj_verify.rb:8:Oj:{: Oj.dump(Math::PI)
```

I came upon this when trying to verify if OJ was being used for `to_json` in a Rails app. Unfortunately I can't think of anything to verify that now, as it looks like `:compat` mode basically locks all of OJ's settings up (my idea was to modify settings on OJ like the time format to see if that changed to_json output).

ohler55:
As far as I know Tracer traces Ruby calls and not C calls. Since Oj is a C extension to Ruby calls internal to Oj will not be traced.

Check out the page on Oj options at http://ohler.com/oj/doc/file.Modes.html. Compat mode is indented to be compatible with the JSON gem which means you do not get options for time formats. The custom mode does provide that feature.

bf4:
Related #460 (really, me an @nateberkopec were chatting about it)

nateberkopec:
> As far as I know Tracer traces Ruby calls and not C calls. Since Oj is a C extension to Ruby calls internal to Oj will not be traced.

Well, I guess that `Tracer.display_c_call = true` isn't very accurate then, is it? It should probably be removed from your docs, then.

ohler55:
It does help identify that Oj is being called just not the internals. @bf4 did a great job updating the docs.

nateberkopec:
I guess I don't understand where in the original output I posted shows that OJ is being called.

ohler55:
Did you try what @bf4 suggested? If not I'll look at it tonight.

[...]

PR #460, opened 2018-02-23
Doc cleanup

bf4:
- Update Rails docs to describe TracePoint usage

- Fix some typos

```bash
misspell  -w -error -source=text -i "clas,class" .
```

I'm not sure if the `clas/class` is important. I didn't look too hard.

There's also an interesting false positive matching 'alse' for false, though I don't
know why the 'f' is missing offhand.

```
+++ b/ext/oj/sparse.c
@@ -161,7 +161,7 @@ read_true(ParseInfo pi) {

 static void
 read_false(ParseInfo pi) {
-    if (0 == reader_expect(&pi->rd, "alse")) {
+    if (0 == reader_expect(&pi->rd, "else")) {
```

### DIFF
--- a/pages/Rails.md
+++ b/pages/Rails.md
@@ -78,7 +78,7 @@ Oj::Rails.optimize is called with no arguments are:
  * any other class where all attributes should be dumped
 
 The ActiveSupport decoder is the JSON.parse() method. Calling the
-Oj::Rails.set_decoder() method replaces that method with the Oj equivelant.
+Oj::Rails.set_decoder() method replaces that method with the Oj equivalent.
 
 ### Notes:
 
--- a/pages/Rails.md
+++ b/pages/Rails.md
@@ -93,6 +93,22 @@ Oj::Rails.set_decoder() method replaces that method with the Oj equivalent.
    are used as keys or if a other non-String objects such as Numerics are mixed
    with numbers as Strings.
 
-3. To verify Oj is being used turn on trace and then set the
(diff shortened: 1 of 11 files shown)


ohler55:
Excellent! I admit I am a horrible speller and not much better at proof reading. Thank you. The extra information on tracing is a help as well.

bf4:
@ohler55 np. Thanks for doing the hard part.  I like https://github.com/client9/misspell and use it in a pre-push hook

pages/Rails.md
```diff
@@ -93,6 +93,22 @@ Oj::Rails.set_decoder() method replaces that method with the Oj equivelant.
    are used as keys or if a other non-String objects such as Numerics are mixed
    with numbers as Strings.
 
-3. To verify Oj is being used turn on trace and then set the
-   `Tracer.display_c_call = true` to see calls to C extensions.
-   
+3. To verify Oj is being used, turn on C extension tracing.
+   Set `tracer = TracePoint.new(:c_call) do |tp| p [tp.lineno, tp.event, tp.defined_class, tp.method_id] end`
+   or, in older Rubies, set `Tracer.display_c_call = true`.
```
bf4: oh, apparently this is in newer rubies too, just need to `require 'tracer'`. Missed that, oops :(

noticed in #459

ohler55: I think folks should be able to figure that much out. :-)

ohler55: Nice hook. I'll start using that too.

PR #479, opened 2018-06-19
Add SemVer compatibility badge to README

greysteil:
First up, thanks for Oj. It's bloody brilliant. 

Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so I threw together the below.

[SemVer compatibility](https://dependabot.com/compatibility-score.html?dependency-name=oj&package-manager=bundler&version-scheme=semver)

If you click through then there's a description of how it's calculated - basically it takes all of the relevant updates Dependabot has done for projects that use `oj` and checks what percentage of the time specs pass on the upgrade PR.

The score is slightly lower than 100% because some projects have flaky specs, but I'm working on filtering them out from the data :octocat:.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # [![{}j](http://www.ohler.com/dev/images/oj_comet_64.svg)](http://www.ohler.com/oj) gem
 
-[![Build Status](https://img.shields.io/travis/ohler55/oj/master.svg)](http://travis-ci.org/ohler55/oj?branch=master) [![AppVeyor](https://img.shields.io/appveyor/ci/ohler55/oj/master.svg)](https://ci.appveyor.com/project/ohler55/oj) ![Gem](https://img.shields.io/gem/v/oj.svg) ![Gem](https://img.shields.io/gem/dt/oj.svg)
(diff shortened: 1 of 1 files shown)


ohler55:
Let me understand a bit better before accepting the PR. I like the idea but was not able to figure out how the score is calculated.

Is the score based on any branch CI failing or just master? Or maybe just tagged releases? Is there a check against the current with the previous unit tests? Is coverage included? Can you explain the formula in more detail?

greysteil:
@ohler55 - it's better than that - it's based on the CI runs of all of Oj's users when upgrading Oj 😎.

Dependabot creates PRs for Oj users whenever a new version is released, and then listens in for whether those PRs pass their CI or not. We use that data to create a "compatibility score" for any given update (e.g., the compatibility score for 3.5.1 -> 3.6.0 is [here](https://dependabot.com/compatibility-score.html?dependency-name=oj&package-manager=bundler&previous-version=3.5.1&new-version=3.6.0)) which we've been including in the PRs for ages (here's an example [PR](https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/307)).

The idea here is to aggregate all of those single-version update scores into an overall SemVer score, which says what percentage of Dependabot PRs pass CI when upgrading between non-major versions.

ohler55:
So not really anything to do with semantic versioning but more a score on how compatible one release is with a previous release, right?

greysteil:
Yep - it won't show whether you're releasing features in patch releases, but it will show if you ever make breaking changes in minor or patch releases (as that would causes people's tests to fail). So it's partially checking SemVer, but that bit is probably the bit that people care about.

ohler55:
Can't hurt to try it out I suppose.

PR #509, opened 2018-10-18
Javascript safe numbers

Ziaw:
Unfortunately, the JavaScript cannot correctly handle numbers larger than Number.MAX_SAFE_INTEGER (9007199254740992).

Greater numbers are not processed correctly, for example, the condition 9007199254740992 == 9007199254740993 is true. JSONs 
{id: 9007199254740992} and {id: 9007199254740993} has equal keys. 

Such numerics can be used as keys if serialized as strings. 

PR adds an option `javascript_safe_numbers` with which numbers greater than 9007199254740992 are dumped as strings.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER


ohler55:
Let me think about it for a day or two. I'm not sure how I feel about linking Oj for Ruby into limitations of Javascript. Maybe there is a more generic solution. Have you tried just setting the `:big decimal_as_decimal` to false?

Ziaw:
>Have you tried just setting the `:big decimal_as_decimal` to false?

Yes, I tried. 9007199254740993 is Fixnum, not BigDecimal. 

Oj excellent tool for send data to Javascript, but this thing does not allow me to use it in full power. My ids has big values.

Appveyor tests failed. I have an idea why and fix it tomorrow.

ohler55:
For a more generic approach how about having an optional limit where a integer type number over or under the negative of that number is written as a string?

Ziaw:
Something like this?
```ruby
javascript_safe_numbers: (-9007199254740992..9007199254740992)
```

ohler55:
I suppose it could be a range. I would choose a different name for the option though. Something that does not say "javascript". Maybe "fixnum_range" or "integer_range". Basically something that indicates the range where integers are written as JSON numbers and not as strings.

Ziaw:
integer_range is ok, but range of Bignum requires ruby interop for bignum comparison. Not good for performance. Any ideas to avoid this?

ohler55:
I think you only care about integers or fixnum, right? BigDecimals are already taken care of with the current options. I assume the range would only been checked if the value being encoded is a fixnum.

Ziaw:
> I assume the range would only been checked if the value being encoded is a fixnum.

OK, it's solution. I will code it tomorrow. In this case fixnum_range is better name, right?

ohler55:
I think so. Maybe not perfect but ok. If you have an alternative that would be okay too.

Thanks for implementing the feature.

Ziaw:
I choose integer_range, fixnum looks is too low-level. Test fail in compat mode on mingw-32. I have no idea why. Can fix it only monday.

Ziaw:
Failed test has been fixed.

ext/oj/dump.c
```diff
@@ -964,16 +964,27 @@ oj_dump_false(VALUE obj, int depth, Out out, bool as_ok) {
 
 void
 oj_dump_fixnum(VALUE obj, int depth, Out out, bool as_ok) {
-    char	buf[32];
+    char	buf[34];
```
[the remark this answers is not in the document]
Ziaw: Change reason is simple: max length increased by 2 chars. I dont know why size was 32. 

32 looks safe for all cases, reverting it.

ext/oj/dump.c
```diff
@@ -997,10 +1013,23 @@ void
 oj_dump_bignum(VALUE obj, int depth, Out out, bool as_ok) {
     volatile VALUE	rs = rb_big2str(obj, 10);
     int			cnt = (int)RSTRING_LEN(rs);
+	bool		dump_as_string = false;
```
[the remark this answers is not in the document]
Ziaw: integer_range_on removed, (out->opts->integer_range_max != 0 || out->opts->integer_range_min != 0) is not good condition for copypaste.

Ziaw:
test failed, fixing it

ohler55:
When you get done with you changes can you squash so it is easier to see what the changes are? Github doesn't make it easy to compare the original to the head of the branch.

ohler55:
No need to do a squash then. From the look of it I thought it was a diff from your previous commit instead of from the master HEAD. 

I still see some of the things I comment on in there. I'll make another pass at it.

ext/oj/oj.c
```diff
@@ -715,6 +736,51 @@ oj_parse_options(VALUE ropts, Options copts) {
 	    }
 	}
     }
+
+    if (Qnil != (v = rb_hash_lookup(ropts, integer_range_sym))) {
+        VALUE is_a_range;
+        VALUE min;
+        VALUE max;
+    
+        switch (TYPE(v))
```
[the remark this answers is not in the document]
ohler55: T_NIL is already checked a few lines earlier. All you care about is if `v` is a `Range` or not. If not then raise else get the min and max. Don't call `is_a?` as it requires a call into Ruby which is more expensive than just `rb_cRange == rb_obj_class(v)`.

ext/oj/oj.h
```diff
@@ -153,6 +153,8 @@ typedef struct _Options {
     char		create_ok;	// YesNo allow create_id
     char		allow_nan;	// YEsyNo for parsing only
     char		trace;		// YesNo
+    int64_t     integer_range_min; // dump numbers outside range as string
```
ohler55: I hate to be picky but please line these up with the rest of the names. Keep the code looking nice.

pages/Modes.md
```diff
@@ -85,7 +85,7 @@ supported in that mode. A number indicates the footnotes describe additional
 information.
 
     | Option                 | type    | :null   | :strict | :compat | :rails  | :object | :custom |  :wab   |
-    | ---------------------- | ------- | ------- | ------- | ------- | ------- | ------- | ------- | ------- |
+    | -----------------------| ------- | ------- | ------- | ------- | ------- | ------- | ------- | ------- |
```
ohler55: Just noticed this. A space is needed between the - and |.

ext/oj/oj.h
Ziaw: Fixed

pages/Modes.md
Ziaw: Fixed

ohler55:
The failing Travis tests are only because travis hasn't provided a 2.5.3 image. They will pass soon or I'll revert to 2.5.1 for travis before making a release. Will be in the air for a day so please give me at least that long to make the release. Thank you for the hard work and new feature.

Ziaw:
Thank you for the merge and constructive dialogue.

PR #534, opened 2019-05-05
Change `Oj.strict_load` to `Oj.safe_load` in Security docs

rafbm:
Security documentation was introduced with the `strict_load` method in 189a372c, but the method was renamed to `safe_load` in 11fd80a9 without updating documentation.


ohler55:
Wow, it took 6 years for that to be noticed. Thanks.

PR #548, opened 2019-07-22
Stop removing `::Parser` top-level constant

paracycle:
This library, as it stands, is removing the top-level `::Parser` constant when the `json.rb` file is required. That breaks the require of other libraries that depend on the `::Parser` constant being available (for example, from https://github.com/whitequark/parser):

```ruby
$ bundle exec pry
[1] pry(main)> require 'parser'
=> true
[2] pry(main)> require 'oj/json'
=> true
[3] pry(main)> require 'rubocop'
NameError: uninitialized constant RuboCop::AST::Parser
Did you mean?  ParseError
from /Users/ufuk/.gem/ruby/2.5.5/gems/rubocop-0.72.0/lib/rubocop/ast/node.rb:21:in `<module:AST>'
```

Instead, the intent here seems to be to just remove the `Parser` constant in the current namespace, which can achieved by doing a simple `remove_const :Parser`.

### DIFF
--- a/lib/oj/json.rb
+++ b/lib/oj/json.rb
@@ -95,7 +95,7 @@ module JSON
   State = ::JSON::Ext::Generator::State unless defined?(::JSON::State)
 
   begin
-    Object.send(:remove_const, :Parser)
+    send(:remove_const, :Parser)
   rescue
   end
   Parser = ::JSON::Ext::Parser unless defined?(::JSON::Parser)


ohler55:
Thanks. I'm surprised this was not caught earlier.

paracycle:
Thank you. I was surprised when I discovered it too. 

Would it be too much to ask you to make a patch release with this fix? A `3.8.1` release perhaps?

ohler55:
Tonight

ohler55:
Released 3.8.1.

PR #552, opened 2019-08-12
Read initial escape_html from active_support in rails mode

artygus:
ActiveSupport configuration usually happens in environment files via app configuration which takes precedence over initializers where `Oj.optimize_rails` supposed to be called. This PR modifies set_encoder method to read initial `escape_html_entities_in_json` value from ActiveSupport


ohler55:
Thanks

Issue #535, opened 2019-05-15
Test with Rails 4.2, 5.2, 6.0 (RC)?


aried3r:
Hey, I just saw that oj is being tested with Rails 4.1.x which is EOL and Rails 5.0.2 specifically. Soon, there will also be Rails 6 released, a RC1 is already released and a `6-0-0-stable` branch cut. I think Rails 4.2 will be EOL once Rails 6 is released, but of course that doesn't mean oj shouldn't support it, unless it requires a lot of additional work.

How does oj's testing strategy work? Personally I'd love to see the currently supported Rails versions being tested, but that might increase test run time a lot, so maybe testing 4.2.x, 5.2.x and 6.0.x should be ok?

And as for Ruby versions, I see [only 2.3 and above](https://github.com/ohler55/oj/blob/06a626f4857cb02548773f88ca850776b2e88b03/.travis.yml#L12-L16) (making [this](https://github.com/ohler55/oj/blob/06a626f4857cb02548773f88ca850776b2e88b03/.travis.yml#L27-L28) obsolete I think) being tested while the `.gemspec` file says [`>= 2`](https://github.com/ohler55/oj/blob/06a626f4857cb02548773f88ca850776b2e88b03/oj.gemspec#L23).

https://weblog.rubyonrails.org/2019/4/27/this-week-in-rails-rails-6-0-0-rc1-is-here-say-hi-at-railsconf/
https://guides.rubyonrails.org/maintenance_policy.html

ohler55:
Thanks, I'll do some CI cleanup this weekend.

ohler55:
I've made the updates to the .travis file as you suggested.

As for testing, Oj has its own tests but also uses the tests from the JSON gem and Rails. When rails is officially released I'll take the same approach with rails 6 and copy the tests and make sure Oj supports them. I anticipate a significant change in Oj if it needs to support both rails 5 and rails 6 as usually rails make breaking changes to the use of JSON. Will have to wait and see.

aried3r:
I see, thank you! Do you plan to keep supporting Rails 4? In which case I'd suggest copying the JSON tests from the latest 4.2 and 5.2 release, what do you think?

ohler55:
I'll most likely drop support for rails 4 when 6 becomes official. That doesn't mean Oj will no longer work with rails 4 but it will no longer be tested against 4.

I should update the tests for rails 5.2 but have not gotten to it yet.

aried3r:
[...]

PR #558, opened 2019-10-03
Use Rails 5.2.3 test cases

aried3r:
This is part of #535. Locally I had a test failure but I hope that this won't be the case on Travis.

Edit: Fixes one test case, but another is failing, which is a new test case that wasn't in 5.0.2 (the previous version of tests).
Here's the bug description in Rails:
https://github.com/rails/rails/issues/26132
And the PR that fixed it:
https://github.com/rails/rails/pull/30953

How would you like to proceed in this case? Is it something that Oj should handle?

### DIFF
--- a/test/activesupport5/decoding_test.rb
+++ b/test/activesupport5/decoding_test.rb
@@ -1,11 +1,13 @@
-require 'activesupport5/test_helper'
-require 'active_support/json'
-require 'active_support/time'
-require 'activesupport5/time_zone_test_helpers'
+# frozen_string_literal: true
 
 require 'oj'
 
 Oj::Rails.set_decoder()
+
+require_relative "abstract_unit"
+require "active_support/json"
+require "active_support/time"
+require_relative "time_zone_test_helpers"
 
 class TestJSONDecoding < ActiveSupport::TestCase
   include TimeZoneTestHelpers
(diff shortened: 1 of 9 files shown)


aried3r:
So the same tests fail on Travis, sadly I have no idea why.

https://travis-ci.org/ohler55/oj/jobs/592965333#L397-L410

```
Failure:
TestJSONEncoding#test_utf8_string_encoded_properly [/home/travis/build/ohler55/oj/test/activesupport5/encoding_test.rb:77]:
Expected: "\"€2.99\""
  Actual: "\"\\u20ac2.99\""
bin/rails test home/travis/build/ohler55/oj/test/activesupport5/encoding_test.rb:75

Failure:
TestJSONEncoding#test_to_json_works_on_io_objects [/home/travis/build/ohler55/oj/test/activesupport5/encoding_test.rb:466]:
Expected: "\"#<IO:0x00000000018d41d0>\""
  Actual: "null"
```

It's this line
https://github.com/ohler55/oj/blob/52affdb19fc49bfb7b181aeb10941106c08e8dd9/test/activesupport5/encoding_test.rb#L77
and this one
https://github.com/ohler55/oj/blob/52affdb19fc49bfb7b181aeb10941106c08e8dd9/test/activesupport5/encoding_test.rb#L466

aried3r:
According to `hexdump`, it's using the correct UTF-8 encoding. But the test case suggests it's UTF-16. I _did_ accidentally remove the `# encoding: UTF-8` header, but I don't think it should matter. At least locally, it doesn't make a difference (the test case still fails with the header).

UTF-8 is `0xE2 0x82 0xAC (e282ac)`
UTF-16 is `0x20AC (20ac)` according to:
https://www.fileformat.info/info/unicode/char/20ac/index.htm

```Console
$ hexdump -C test/activesupport5/encoding_test.rb | rg "e2 82[[:space:]]+ac" -A2
00000a90  2e 65 6e 63 6f 64 65 28  22 e2 82 ac 32 2e 39 39  |.encode("...2.99|
00000aa0  22 29 0a 20 20 20 20 61  73 73 65 72 74 5f 65 71  |").    assert_eq|
00000ab0  75 61 6c 20 27 22 e2 82  ac 32 2e 39 39 22 27 2c  |ual '"...2.99"',|
00000ac0  20 72 65 73 75 6c 74 0a  20 20 20 20 61 73 73 65  | result.    asse|
00000ad0  72 74 5f 65 71 75 61 6c  28 45 6e 63 6f 64 69 6e  |rt_equal(Encodin|
```

However, Ruby thinks this of the string. Encoding is UTF-8 but it prints the UTF-16 representation in hexadecimal. (20AC)
```Ruby
result = ActiveSupport::JSON.encode("€2.99")
pp result.encoding  #<Encoding:UTF-8>
pp "€2.99".encoding #<Encoding:UTF-8>
pp "€2.99".chars.map { |c| puts "%s %3d %02X" % [ c, c.ord, c.ord ] }
# € 8364 20AC
# 2  50 32
# .  46 2E
# 9  57 39
# 9  57 39
```

The same happens if I force the encoding using:
```Ruby
"€2.99".dup.force_encoding('UTF-8')
```

aried3r:
Of course the answer is much easier. I was just not careful enough copying the test cases.

I was missing:
```Ruby
    # The original test seems to expect that
    # ActiveSupport.escape_html_entities_in_json reverts to true even after
    # being set to false. I haven't been able to figure that out so the value is
    # set to true, the default, before running the test. This might be wrong but
    # for now it will have to do.
    ActiveSupport.escape_html_entities_in_json = true
```

test/activesupport5/decoding_test.rb
```diff
@@ -1,12 +1,14 @@
...
+# frozen_string_literal: true
 
 require 'oj'
 
 Oj::Rails.set_decoder()
 
+require_relative "abstract_unit"
+require "active_support/json"
+require "active_support/time"
+require_relative "time_zone_test_helpers"
```
ohler55: I think the Oj code needs to be after this or else the active support decoder is used. Have you checked?

aried3r: Ah, that's a good point! I do remember a way to check if Oj is being used, but can't find it in the docs atm. Could you maybe point me into the right direction? Then I can also add a test case for it :)

test/activesupport5/encoding_test.rb
```diff
@@ -454,15 +460,19 @@ def test_to_json_works_when_as_json_returns_infinite_number
...
   end
 
   def test_to_json_works_when_as_json_returns_NaN_number
     assert_equal '{"number":null}', NaNNumber.new.to_json
   end
 
-  protected
+  def test_to_json_works_on_io_objects
+    assert_equal STDOUT.to_s.to_json, STDOUT.to_json
+  end
```
aried3r: This is the one failing test case btw. It's described in the OP of this PR.

test/activesupport5/decoding_test.rb
ohler55: You can turn on tracing but that doesn't help for testing. How about checking the `ActiveSupport::json_encoder` or is it `ActiveSupport.json_encoder`.

test/activesupport5/encoding_test.rb
ohler55: I think that might be a change in behaviour or maybe the test was not in the older version of the test suite. Anyway, Oj return `null` for IO objects. I'd have to look into the code to determine why. I can do that later this evening.

ohler55: I see. In rails.c:1442 there is no encoder for file types. Rails wants a string.

ohler55: replace that line with this

```
    dump_as_string,	// RUBY_T_FILE     = 0x0b,
```

test/activesupport5/decoding_test.rb
aried3r: Hmm, I can test this in the **en**coding test, but there is no `json_decoder`, so the test fails (I added it anyway, to see what I tried).

test/activesupport5/decoding_test.rb
```diff
@@ -10,83 +12,90 @@
 class TestJSONDecoding < ActiveSupport::TestCase
   include TimeZoneTestHelpers
 
+  # Added for testing if Oj is used.
+  test "oj is used as an encoder" do
+    assert_equal ActiveSupport.json_encoder, Oj::Rails::Encoder
+  end
```
aried3r: This doesn't check for a decoder, but the encoder. I'm not sure if that actually tests the proper thing, since there is no `json_decoder` or `json_parser`.

ohler55: Probably good enough. They are both set at the same time.

ohler55:
Thanks for the help getting Oj to pass the more current tests.

PR #566, opened 2019-10-31
Travis: Use the latest ruby patch versions

sdemjanenko:
We should be running tests against the latest and greatest ruby patches
for each minor version that this library supports.


ohler55:
Beat me to it. Thanks.

PR #578, opened 2020-01-29
Rails compatibility documentation update

rudolfolah:
- how to use Oj with Rails 3
- which version of Oj to use with earlier Ruby (pre 2.3.0) versions


ohler55:
Thanks for the updates.

PR #580, opened 2020-02-10
Rails.md: Rails 3 encoder example update

rudolfolah:
- code updated to properly handle already encoded JSON strings


ohler55:
Thanks for the contribution.

PR #589, opened 2020-03-13
fix API doc

paddor:
My IDE (RubyMine) keeps complaining about a missing argument to `Oj.dump` and `Oj.load`. This should fix some or all of the `call-seq` occurrences.


ohler55:
Thanks for the help.

PR #595, opened 2020-04-27
Update Readme

olkinn:


ohler55:
I haven't made that change in the past since rdoc needed the old format. Have you tried to see if links work with rdoc?

ohler55:
If you can figure out how to get the file links to look good in both yardoc output and GitHub that would be great. I have't figured out the magic. As it is, the change you propose breaks the yardoc pages. The existing doesn't look great on GitHub but it does work.

olkinn:
@ohler55 I haven't had the time to look into those services unfortunately :/

haines:
I've run into that problem with links in the past, and ended up writing a YARD plugin to solve it: https://github.com/haines/yard-relative_markdown_links

PR #599, opened 2020-05-13
run on latest Ruby runtimes

mt-kelvintaywl:
#### description

There has been an update on the Ruby releases since March 31.
https://www.ruby-lang.org/en/downloads/releases/

- 2.4.10
- 2.5.8
- 2.6.6
- 2.7.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,10 @@ env:
 language: ruby
 rvm:
   - 2.3.8
-  - 2.4.9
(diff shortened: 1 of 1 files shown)


ohler55:
I would be glad to bump the versions up for Oj CI. Have to wait for Travis to catch up though. Back off until travis has the versions or work around that in some other way and I'd be happy to merge this.

PR #618, opened 2020-08-26
fix fairly significant typo

ghiculescu:


ohler55:
Yup, pretty significant. Thanks.

PR #625, opened 2020-11-09
Enabled wider compatibility for .compat_load, better parity with JSON.parse in Ruby

MarkYHZhang:
Greater parity with `JSON.parse` on ruby, where it will ignore `/` in case of an invalid escape character. Such as `\”` (curly double quote) will be parsed to just `”`

### DIFF
--- a/ext/oj/parse.c
+++ b/ext/oj/parse.c
@@ -212,18 +212,6 @@ read_escaped_str(ParseInfo pi, const char *start) {
 	    case '"':	buf_append(&buf, '"');	break;
 	    case '/':	buf_append(&buf, '/');	break;
 	    case '\\':	buf_append(&buf, '\\');	break;
-	    case '\'':
-		// The json gem claims this is not an error despite the
-		// ECMA-404 indicating it is not valid.
-		if (CompatMode == pi->options.mode) {
-		    buf_append(&buf, '\'');
-		} else {
-		    pi->cur = s;
-		    oj_set_error_at(pi, oj_parse_error_class, __FILE__, __LINE__, "invalid escaped character");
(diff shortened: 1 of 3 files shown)


ext/oj/parse.c
```diff
@@ -263,6 +263,10 @@ read_escaped_str(ParseInfo pi, const char *start) {
 		}
 		break;
 	    default:
+		if (CompatMode == pi->options.mode) {
+		    buf_append(&buf, *s);
+		    break;
+		}
```
pushrax, line 269: Can remove the special case for `\'` now. Should also add a similar comment here.


	// The json gem claims this is not an error despite the
	// ECMA-404 indicating it is not valid.

pushrax (commented):
There is another parser that supports compat mode, which also should be changed for consistency. https://github.com/ohler55/oj/blob/1a1c155e3b83d10d1fe74f7468e1d2a9ef332a2a/ext/oj/sparse.c#L275-L278

I'll add that we can see the Ruby JSON implementation allows any character to be escaped in https://github.com/ruby/ruby/blob/cbe869b3707feac4e3de22ab63c40fb881640c53/ext/json/parser/parser.c#L1433-L1509

ohler55:
Lets see what travis says. I expect those CI runs to pass though. I'll merge when it does.

MarkYHZhang:
Hey thanks for merging it, I wonder what's the timeline looking like for merging this into `master`? We're hoping to get the latest version with this fix to address some of the issues we have.

pushrax:
We can use the git ref of the develop branch until then @MarkYHZhang .

ohler55:
Released!

PR #638, opened 2021-01-23
Fix object movement bugs

### LINKED ISSUE #637: GC.compact moves references held by Oj, leading to TypeErrors
After loading Oj and running GC.compact, errors like this can occur ``... in `safe_load': wrong argument type Time (expected Class) (TypeError)``

@casperisfine worded this well:
> Since Ruby 2.7, C global variables holding references to Ruby objects must be declared to the GC even if they are also held from the Ruby side, otherwise a call to GC.compact might move them to another location.

Examples of fixes to other C extensions:

https://github.com/brianmario/mysql2/pull/1115
https://github.com/Shopify/semian/pull/275

cc @XrXr @casperisfine this is blocking use of GC.compact in SFR, if you feel like tackling it. Or @ohler55 of course 😄

XrXr:
Fixes #637

### DIFF
--- a/ext/oj/mimic_json.c
+++ b/ext/oj/mimic_json.c
@@ -841,6 +841,7 @@ oj_mimic_json_methods(VALUE json) {
     }
     // Pull in the JSON::State mimic file.
     state_class = rb_const_get_at(generator, rb_intern("State"));
+    rb_gc_register_mark_object(state_class);
 
     symbolize_names_sym = ID2SYM(rb_intern("symbolize_names"));	rb_gc_register_address(&symbolize_names_sym);
 }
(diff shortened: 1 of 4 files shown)


ohler55:
Is that all that's needed. Nice. You are a star. I'll merge once travis completes.

XrXr:
Thanks! On a somewhat related note, it might be a good idea to add a CI check for Ruby 3.0.0.

ohler55:
I tried a week or 2 ago and travis still don't support 3.0. Maybe time to try again. Travis always lags on new Ruby releases.

ohler55:
Travis still does not support Ruby 3.0.0 on macOS.

PR #642, opened 2021-01-28
Update CHANGELOG.md

nicovak:
New changelog entries should be in 2021 now


ohler55:
Thanks for catching that.

PR #647, opened 2021-03-30
Remove duplicates of Date

randsina:
The serializer of `Date` in the mimic duplicates.
https://github.com/ohler55/oj/blob/620a787135ef9c1937e2265af7b11cace448038a/lib/oj/mimic.rb#L136-L146
https://github.com/ohler55/oj/blob/620a787135ef9c1937e2265af7b11cace448038a/lib/oj/mimic.rb#L177-L187

### DIFF
--- a/lib/oj/mimic.rb
+++ b/lib/oj/mimic.rb
@@ -133,18 +133,6 @@ module Oj
       end
     end
 
-    Date.class_eval do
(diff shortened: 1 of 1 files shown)


ohler55:
Thanks. Funny how long it's been like that and no one else caught it.

PR #648, opened 2021-03-31
Install rails dependency for osx in Travis

randsina:
Update Rails versions to skip broken `mimemagic` gem and skip `shared-mime-info` native extension installation.
https://github.com/rails/rails/issues/41750


ohler55:
Pending CI passing it looks fine.

PR #653, opened 2021-04-09
Make compaction available for `Oj::Doc`

BuonOmo:
Introduce compaction to `fast.c`, only if `rb_gc_mark_movable` is
available. Otherwise, keep a version that supports ruby <2.7.

Since `rb_data_object_wrap` and `TypedData_Wrap_Struct` is available
in every supported versions (eg >=2.4.10), I've also removed support
for earlier versions.

Result for compaction is great. I've tested it by allocating and
removing arbitrary sized json. For a count of 70 pages, we can see
that we used to need 68 pages after compaction,and now 61 pages. Of
course this test include also lots of noise, but if we count pinned
objects (objects that are marked in C extension as uncollectible),
we can see that for a benchmark with no allocations, we have the
same amount as a benchmark with this commit with thousands of allocations
(499 pinned objects total), and before this, `Oj::Doc` used to generate
5k more objects (5499 pinned objects total).

Hence this is a win, see ohler55/oj#650 for a graphical result.

Signed-off-by: Ulysse Buonomo <buonomo.ulysse@gmail.com>


test/test_scp.rb
```diff
@@ -328,7 +328,7 @@ def test_pipe
     IO.pipe do |read_io, write_io|
       if fork
         write_io.close
-        Oj.sc_parse(handler, read_io) {|v| p v}
+        Oj.sc_parse(handler, read_io)
```
BuonOmo, line 331: @ohler55 this one was outputing to tests, I looks like this was unwanted. Not 100% sure though

BuonOmo:
I've also renamed the `clang-format` file [according to documentation](https://releases.llvm.org/11.0.0/tools/clang/docs/ClangFormatStyleOptions.html) 🙂

ohler55 (approved):

ohler55:
Thanks for the additions and flexibility in getting it in.

PR #657, opened 2021-04-28
Switch from travis to github actions

BuonOmo:
Since travis is saying goodbye to Open-Source, I've switched to Github Actions.

I've tried to include Windows but:
- every tests are failing,
- it goes from 3mn to 15mn to finish the CI run.

I've dropped ruby 2.4 support since it is EOL.

Apart from that choice I sticked as much as possible to the travis config. You can see the result on my fork: https://github.com/BuonOmo/oj/actions

@ohler55 on your part you will have to update branch protection rules in the settings.

Tell me if something needs changes :)
I can squash as well


ohler55:
I can squash as part of the merge. No problem there. Not having Windows is fine too. Let me know what needs to change on the protection rules and I'll merge and make the changes.

BuonOmo:
@ohler55 I don't know since I do not have access to your repository, but the branch protection rules are the one which can prevent a merge if there is a test failure. If you are the sole maintainer, maybe there is no need for any rule !

ohler55:
Right now I am the sole maintainer. It's worth a shot merging I think. Worse case is some debugging of the process is required.

ohler55:
Works like a charm. Thank you very much.

PR #670, opened 2021-07-06
Cache

ohler55:
Adds hash key caching as well as string caching. Both are options. Gives a 20% performance boost.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 3.12.0 - 2021-07-05
+
+- Added string and symbol caching options that give Oj about a 20% parse performance boost.
+
 ## 3.11.8 - 2021-07-03
 
 - Fixed or reverted change that set the default mode when optimize_Rails was called.
--- a/ext/oj/compat.c
+++ b/ext/oj/compat.c
@@ -23,14 +23,26 @@ static void hash_set_cstr(ParseInfo pi, Val kval, const char *str, size_t len, c
         parent->classname = oj_strndup(str, len);
         parent->clen      = len;
     } else {
-        volatile VALUE rstr = rb_str_new(str, len);
+        volatile VALUE rstr = oj_cstr_to_value(str, len, (size_t)pi->options.cache_str);
 
         if (Qundef == rkey) {
-            rkey = rb_str_new(key, klen);
-            rstr = oj_encode(rstr);
-            rkey = oj_encode(rkey);
+            VALUE *slot;
+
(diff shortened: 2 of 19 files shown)


PikachuEXE:
Will the new options `cache_str` & `cache_keys` be documented?

ohler55:
I guess I should upload the new docs. :-)

ohler55:
Ok, updated the docs on the web site.

xokocodo:
Hi,

ohler55:
If using many keys then it would be best if `cache_keys` was turned off. When turned off the performance should be the same as previous versions. Are you seeing something different?

xokocodo:
Yeah, with some large JSON files I have I am seeing a 10x running time versus version `3.11.8`, even with `cache_keys` turned off. I haven't looked into it too deeply but just wanted to check if this was a known issue or if there was something I am missing.

ohler55:
Do you have a file I can test against?

xokocodo:
Sure, here is a link to one of the files (19MB): https://drive.google.com/file/d/10sZ-8ksIfgDsq-kfPJUECa9WffzXx8u5/view?usp=sharing

In my environment this takes ~2.4 seconds with `3.11.8` and ~25 seconds with `3.12.0`

ohler55:
Thank you. What mode are you using? `:compat`, `:object`, etc

I'll dig into it tomorrow and get it resolved one way or another.

xokocodo:
Thank you! I was using the default (`:object`) mode but it looks like the difference is consistent across different modes

ohler55:
Pushed a "cache-fix" branch. The primary fix was to have `:compat` mode honor the `:cache_keys` option. I suspect that was the mode you were using. `:object` mode does not use the cache and `:strict` was already honoring the option.

The secondary change was to increase the cache size. Being a cache it will only show any kind of improvement with repetitions of the keys in either the same file or in multiple files. Even then the number of different keys should be less than something like 100,000. Maybe lower like 20,000. It really depends on the situation and benchmarking with the option on and off would be worthwhile for any given situation.

Anyway, please let me know what you think about the branch.

xokocodo:
The branch seems to be performing much better again (back around ~3 seconds). I tried a few different modes again and they are all much faster. I'm not really sure why that would be the case, but maybe I'm doing something wrong.

Anyways this resolves the issue I was seeing. Thank you very much for taking a look at it!

ohler55:
For that JSON and only parsing it one time I would not turn caching on. Each key is unique and there are around 700,000 keys. Far outside the recommended limits.

Thanks for bringing the issue to my attention.

PR #671, opened 2021-07-13
Make sure Oj.default_options hash is unable to be modified directly

semaperepelitsa:
Since modifying the hash is not supported, we can freeze it to avoid possible mistakes.

> Attempting to modify the Oj.default_options Hash directly will not set the changes on the actual default options but on a copy of the Hash

### DIFF
--- a/ext/oj/oj.c
+++ b/ext/oj/oj.c
@@ -499,7 +499,7 @@ static VALUE get_def_opts(VALUE self) {
         }
         rb_hash_aset(opts, ignore_sym, a);
     }
-    return opts;
+    return rb_obj_freeze(opts);
 }
 
 /* Document-method: default_options=
--- a/test/test_various.rb
+++ b/test/test_various.rb
@@ -152,6 +152,10 @@ class Juice < Minitest::Test
     assert_equal(orig, opts);
   end
 
+  def test_options_frozen
+    assert_raises(FrozenError) { Oj.default_options[:mode] = :strict }
+  end
+
   def test_nil
     dump_and_load(nil, false)
   end


ohler55:
I'm not sure why you think the options hash should be frozen. Modifying the hash has no effect on Oj or the default options. Is there a reason to freeze the hash? If a use wants to take a copy, modify and then set with the modified hash I see no reason to stop them.

semaperepelitsa:
Sorry if I wasn't clear. Previously, I was trying to change the default options and got confused why it wasn't working. 

```ruby
Oj.default_options[:mode] = :strict
```

Only after reading deeper into the documentation I realised how you are supposed to do it.

```ruby
Oj.default_options = {:mode => :strict}
```

Would you agree that this behaviour is not very intuitive for an average Ruby developer who will use this library? If so, this small change will help avoid that confusion, with an explicit exception raised.

> take a copy, modify and then set with the modified hash

I haven't thought about that use case because setting the default options like above doesn't actually override any existing options, so it would be pointless to modify the hash and set it back when you can set the new options directly. But if you have concerns about backwards compatibility, I understand.

semaperepelitsa:
P. S. I saw this technique used in other Ruby libraries, that's where I got this idea.

ohler55:
I would agree that the approach Oj uses is not as intuitive. It does need to remain in place even if another approach is also provided for backward compatibility. If you would like to implement the more intuitive approach I'd be glad to review the MR. If not then open an issue and I can address that myself.

I am actually exploring a third options with a new parser based on what I did with OjC. It is providing to be around 15% faster but it is in the early stages. On that parser the options are set directly on the parser itself so much more intuitive.

ohler55:
I'm going to close this MR. If the MR was to add support for the Oj.default_options[:xxx] = yyy then I'd consider but just freezing the return doesn't provide any real benefit.

PR #674, opened 2021-07-25
Improve `Oj.dump` performance

Watson1978:
This patch uses standard C library to copy the string
because copying one byte at a time is slow.

This patch will improve `Oj.dump` performance as following.

–                | before   | after    | result
Oj.dump          | 689.236k | 1.853M   | 2.69x
Oj.dump (compat) | 476.107k | 827.446k | 1.74x
Oj.dump (rails)  | 464.545k | 644.494k | 1.39x

### Environment
- MacBook Air (M1, 2020)
- macOS 12.0 beta 3
- Apple M1
- Ruby 3.0.2

### Before
```
             Oj.dump    69.210k i/100ms
    Oj.dump (compat)    47.123k i/100ms
     Oj.dump (rails)    45.911k i/100ms
             Oj.dump    689.236k (± 0.2%) i/s -      3.460M in   5.020801s
    Oj.dump (compat)    476.107k (± 0.9%) i/s -      2.403M in   5.048128s
     Oj.dump (rails)    464.545k (± 0.9%) i/s -      2.341M in   5.040711s
```

### After
```
             Oj.dump   187.096k i/100ms
    Oj.dump (compat)    82.879k i/100ms
     Oj.dump (rails)    64.371k i/100ms
             Oj.dump      1.853M (± 0.3%) i/s -      9.355M in   5.049406s
    Oj.dump (compat)    827.446k (± 0.2%) i/s -      4.144M in   5.008145s
     Oj.dump (rails)    644.494k (± 0.2%) i/s -      3.283M in   5.093814s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

data = {
  'short_string': 'a' * 50,
  'long_string': 'b' * 255,
  'utf8_string': 'あいうえお' * 10
}

Benchmark.ips do |x|
  x.report('Oj.dump (compat)') { Oj.dump(data, mode: :compat) }
  x.report('Oj.dump (rails)') { Oj.dump(data, mode: :rails) }
end
```

### DIFF
--- a/ext/oj/dump.c
+++ b/ext/oj/dump.c
@@ -535,7 +535,7 @@ void oj_dump_xml_time(VALUE obj, Out out) {
     }
     if ((0 == nsec && !out->opts->sec_prec_set) || 0 == out->opts->sec_prec) {
         if (0 == tzsecs && rb_funcall2(obj, oj_utcq_id, 0, 0)) {
-            sprintf(buf,
+            int len = sprintf(buf,
                     "%04d-%02d-%02dT%02d:%02d:%02dZ",
                     ti.year,
                     ti.mon,
@@ -543,9 +543,9 @@ void oj_dump_xml_time(VALUE obj, Out out) {
                     ti.hour,
                     ti.min,
                     ti.sec);
-            oj_dump_cstr(buf, 20, 0, 0, out);
+            oj_dump_cstr(buf, len, 0, 0, out);
         } else {
-            sprintf(buf,
+            int len = sprintf(buf,
                     "%04d-%02d-%02dT%02d:%02d:%02d%c%02d:%02d",
                     ti.year,
                     ti.mon,
@@ -556,27 +556,25 @@ void oj_dump_xml_time(VALUE obj, Out out) {
                     tzsign,
                     tzhour,
                     tzmin);
-            oj_dump_cstr(buf, 25, 0, 0, out);
+            oj_dump_cstr(buf, len, 0, 0, out);
         }
     } else if (0 == tzsecs && rb_funcall2(obj, oj_utcq_id, 0, 0)) {
         char format[64] = "%04d-%02d-%02dT%02d:%02d:%02d.%09ldZ";
-        int  len        = 30;
+        int  len;
 
         if (9 > out->opts->sec_prec) {
             format[32] = '0' + out->opts->sec_prec;
-            len -= 9 - out->opts->sec_prec;
         }
-        sprintf(buf, format, ti.year, ti.mon, ti.day, ti.hour, ti.min, ti.sec, (long)nsec);
+        len = sprintf(buf, format, ti.year, ti.mon, ti.day, ti.hour, ti.min, ti.sec, (long)nsec);
         oj_dump_cstr(buf, len, 0, 0, out);
     } else {
         char format[64] = "%04d-%02d-%02dT%02d:%02d:%02d.%09ld%c%02d:%02d";
-        int  len        = 35;
+        int  len;
 
         if (9 > out->opts->sec_prec) {
             format[32] = '0' + out->opts->sec_prec;
-            len -= 9 - out->opts->sec_prec;
         }
-        sprintf(buf,
+        len = sprintf(buf,
                 format,
                 ti.year,
                 ti.mon,
@@ -827,9 +825,8 @@ void oj_dump_cstr(const char *str, size_t cnt, bool is_sym, bool escape1, Out ou
         if (is_sym) {
             *out->cur++ = ':';
         }
-        for (; '\0' != *str; str++) {
-            *out->cur++ = *str;
-        }
+        strncpy(out->cur, str, cnt);
+        out->cur += cnt;
         *out->cur++ = '"';
     } else {
         const char *end         = str + cnt;


Watson1978:
I retrieved a benchmark result on Intel Mac.

## Env
- MacBook Pro (16-inch, 2019)
- macOS BigSur 11.5
- CPU 2.4 GHz 8cores Intel Core i9

## Before

```
             Oj.dump    22.840k i/100ms
    Oj.dump (compat)     7.228k i/100ms
     Oj.dump (rails)     7.017k i/100ms
             Oj.dump    205.030k (± 5.9%) i/s -      1.028M in   5.030663s
    Oj.dump (compat)     83.240k (± 4.5%) i/s -    419.224k in   5.047241s
     Oj.dump (rails)     71.655k (±10.1%) i/s -    357.867k in   5.077355s
```

## After
```
             Oj.dump    27.699k i/100ms
    Oj.dump (compat)     9.134k i/100ms
     Oj.dump (rails)     8.325k i/100ms
             Oj.dump    278.205k (± 3.8%) i/s -      1.413M in   5.085761s
    Oj.dump (compat)     92.628k (± 3.9%) i/s -    465.834k in   5.037398s
     Oj.dump (rails)     81.348k (± 4.6%) i/s -    407.925k in   5.025928s
```

ext/oj/dump.c
```diff
@@ -535,17 +535,17 @@ void oj_dump_xml_time(VALUE obj, Out out) {
     }
     if ((0 == nsec && !out->opts->sec_prec_set) || 0 == out->opts->sec_prec) {
         if (0 == tzsecs && rb_funcall2(obj, oj_utcq_id, 0, 0)) {
-            sprintf(buf,
+            int len = sprintf(buf,
```
ohler55, line 538: This change really isn't needed. I doubt the difference is enough to measure but creating a variable for something that will always be the same value has to be more expensive than a constant.

ext/oj/dump.c
```diff
@@ -535,17 +535,17 @@ void oj_dump_xml_time(VALUE obj, Out out) {
...
                     ti.mon,
                     ti.day,
                     ti.hour,
                     ti.min,
                     ti.sec);
-            oj_dump_cstr(buf, 20, 0, 0, out);
+            oj_dump_cstr(buf, len, 0, 0, out);
         } else {
-            sprintf(buf,
+            int len = sprintf(buf,
```
ohler55, line 548: Same here, no need for a variable when the value will always be the same.

ext/oj/dump.c
```diff
@@ -566,7 +566,7 @@ void oj_dump_xml_time(VALUE obj, Out out) {
             format[32] = '0' + out->opts->sec_prec;
             len -= 9 - out->opts->sec_prec;
         }
-        sprintf(buf, format, ti.year, ti.mon, ti.day, ti.hour, ti.min, ti.sec, (long)nsec);
+        len = sprintf(buf, format, ti.year, ti.mon, ti.day, ti.hour, ti.min, ti.sec, (long)nsec);
```
ohler55, line 569: I'm not sure which is best for determining the length. I suspect you change where Len is assigned to the return of sprintf is best which means the line above `len -= 9 - out->opts->sec_prec;` can be removed. No need to set Len to 30 either if using the sprintf return value.

ext/oj/dump.c
```diff
@@ -576,7 +576,7 @@ void oj_dump_xml_time(VALUE obj, Out out) {
             format[32] = '0' + out->opts->sec_prec;
             len -= 9 - out->opts->sec_prec;
         }
-        sprintf(buf,
+        len = sprintf(buf,
```
ohler55, line 579: Same on this change. Best to remove the earlier calculation of Len since it is over-ridden by this line.

ohler55:
I made a few comments but please don't take that the wrong way. Your catch on the use of strncpy was a good one. I'd be curious if memcpy would be a bit better. Would you mind running the benchmarks to compare?

As for the other comments, I'd like to keep the constants on the ones that always return the same length but the two that use the sprintf return value are keepers so it is better to remove the non-used calculations of len earlier in the code.

ohler55 (changes requested):

ext/oj/dump.c
Watson1978: Unfortunately, without this change, the test will fail.

```
$ rake
mkdir -p tmp/arm64-darwin21/oj/3.0.2
cd tmp/arm64-darwin21/oj/3.0.2
/Users/watson/.rbenv/versions/3.0.2/bin/ruby -I. ../../../../ext/oj/extconf.rb
>>>>> Creating Makefile for ruby version 3.0.2 on arm64-darwin21 <<<<<
checking for rb_time_timespec()... yes
checking for rb_ivar_count()... yes
checking for rb_ivar_foreach()... yes
checking for rb_gc_mark_movable()... yes
checking for stpcpy()... yes
checking for pthread_mutex_init()... yes
creating Makefile
cd -
cd tmp/arm64-darwin21/oj/3.0.2
/usr/bin/make
compiling ../../../../ext/oj/cache8.c
compiling ../../../../ext/oj/circarray.c
compiling ../../../../ext/oj/code.c
compiling ../../../../ext/oj/compat.c
compiling ../../../../ext/oj/custom.c
compiling ../../../../ext/oj/dump.c
compiling ../../../../ext/oj/dump_compat.c
compiling ../../../../ext/oj/dump_leaf.c
compiling ../../../../ext/oj/dump_object.c
compiling ../../../../ext/oj/dump_strict.c
compiling ../../../../ext/oj/err.c
compiling ../../../../ext/oj/fast.c
compiling ../../../../ext/oj/hash.c
compiling ../../../../ext/oj/hash_test.c
compiling ../../../../ext/oj/mimic_json.c
compiling ../../../../ext/oj/object.c
compiling ../../../../ext/oj/odd.c
compiling ../../../../ext/oj/oj.c
compiling ../../../../ext/oj/parse.c
compiling ../../../../ext/oj/rails.c
compiling ../../../../ext/oj/reader.c
compiling ../../../../ext/oj/resolve.c
compiling ../../../../ext/oj/rxclass.c
compiling ../../../../ext/oj/saj.c
compiling ../../../../ext/oj/scp.c
compiling ../../../../ext/oj/sparse.c
compiling ../../../../ext/oj/stream_writer.c
compiling ../../../../ext/oj/strict.c
compiling ../../../../ext/oj/string_writer.c
compiling ../../../../ext/oj/trace.c
compiling ../../../../ext/oj/util.c
compiling ../../../../ext/oj/val_stack.c
compiling ../../../../ext/oj/wab.c
linking shared-object oj/oj.bundle
cd -
mkdir -p tmp/arm64-darwin21/stage/lib/oj
install -c tmp/arm64-darwin21/oj/3.0.2/oj.bundle lib/oj/oj.bundle
cp tmp/arm64-darwin21/oj/3.0.2/oj.bundle tmp/arm64-darwin21/stage/lib/oj/oj.bundle

##########################################################################################
ruby test/tests.rb && ruby test/tests_mimic.rb && ruby test/tests_mimic_addition.rb
Run options: --seed 31807

# Running:

.......................................................................................F................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 0.026347s, 18522.0329 runs/s, 34842.6766 assertions/s.

  1) Failure:
Juice#test_time_years [/Users/watson/prj/oj/test/test_various.rb:407]:
Expected: "-2020-03-01T15:14:13Z"
  Actual: "-2020-03-01T15:14:13"

488 runs, 918 assertions, 1 failures, 0 errors, 0 skips

....
```

Before, I think the wong length was passed into ob_dump_str().

ext/oj/dump.c
Watson1978: Same above reply

ext/oj/dump.c
Watson1978: Thanks. I removed redundant codes.

ext/oj/dump.c
Watson1978: Thanks. I removed redundant codes.

ext/oj/dump.c
ohler55: That is an interesting observation. RFC3339 doesn't support negative years although it does make sense to so I'd have to agree.

ohler55:
Thank you for the nice improvement. That will go nicely with the parser re-write I'm in the middle of but I'll release your improvement sooner than the parser.

PR #690, opened 2021-08-11
Fix typos

kamipo:


ohler55:
There were quite a few spelling errors. Thanks although the "Hashi" test name was not a spelling error. Just being cute.

kamipo:
Ah I was also wondering if "Hashi" was misspelled, but I'd followed "HashTest" in the end comment.

https://github.com/ohler55/oj/blob/38be7638f2933b123d4e67ac8f40da06617e346b/test/test_hash.rb#L39

ohler55:
Probably went through a change or two. All good though. Thank you.

Issue #689, opened 2021-08-10
Verison 3.13.x crashes cmake - error in cache.c


RemcoVanDerVeen:
Updating oj to version 3.13.0 resulted in the following error in our deploy cycle. It was not solved by 3.13.1.
Here is the error log:

```
current directory: /usr/local/rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/oj-3.13.1/ext/oj
/usr/local/rbenv/versions/2.6.1/bin/ruby -I /usr/local/rbenv/versions/2.6.1/lib/ruby/site_ruby/2.6.0 -r ./siteconf20210810-49232-alhbhg.rb extconf.rb
>>>>> Creating Makefile for ruby version 2.6.1 on x86_64-linux <<<<<
checking for rb_time_timespec()... yes
checking for rb_ivar_count()... yes
checking for rb_ivar_foreach()... yes
checking for rb_gc_mark_movable()... no
checking for stpcpy()... yes
checking for pthread_mutex_init()... yes
checking for rb_enc_associate()... yes
checking for rb_ext_ractor_safe() in ruby.h... no
creating Makefile

current directory: /usr/local/rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/oj-3.13.1/ext/oj
make DESTDIR\= clean

current directory: /usr/local/rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/oj-3.13.1/ext/oj
make DESTDIR\=
compiling cache.c
cache.c: In function ‘rehash’:
cache.c:115:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (Slot *sp = c->slots; sp < end; sp++) {
     ^
cache.c:115:5: note: use option -std=c99 or -std=gnu99 to compile your code
cache.c: In function ‘cache_free’:
cache.c:133:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (uint32_t i = 0; i < c->size; i++) {
     ^
cache.c:135:9: error: ‘for’ loop initial declarations are only allowed in C99 mode
         for (Slot s = c->slots[i]; NULL != s; s = next) {
         ^
cache.c: In function ‘cache_mark’:
cache.c:146:9: error: ‘for’ loop initial declarations are only allowed in C99 mode
         for (uint32_t i = 0; i < c->size; i++) {
         ^
cache.c:147:13: error: ‘for’ loop initial declarations are only allowed in C99 mode
[...]

PR #692, opened 2021-08-12
Make code compatible with C90

andrepiske:
Trying to get #689 through.

I was running `gem build ...` and `gem install ...` and fixing all compiler errors regarding C90 compatibility until it worked.


ohler55 (approved):

ohler55:
Thank you!

PR #694, opened 2021-08-15
Improve `Oj.dump` performance with `Time` object

Watson1978:
If `Time#to_json` existed, it would have been called.
And `Time#to_json` is quite slow.

However, if you just call `require 'json'` before, it would define `Time#to_json`.
`Time#to_json` and `Time#to_s` have same behavior until overriding `Time#to_json` by calling `require 'json/add/time'`.

If you really want to use `Time#to_json`, you can use `use_to_json` flag.
`Oj.dump(Time.now, mode: :compat, use_to_json: true)`

Oj.dump          | 4.454M   | 4.452M   | -
Oj.dump (compat) | 585.849k | 1.038M   | 1.77x
Oj.dump (rails)  | 1.012M   | 1.006M   | -

### Environment
- MacBook Air (M1, 2020)
- macOS 12.0 beta 4
- Apple M1
- Ruby 3.0.2

### Before
```
       JSON.generate    61.732k i/100ms
             Oj.dump   452.219k i/100ms
    Oj.dump (compat)    59.191k i/100ms
     Oj.dump (rails)   101.236k i/100ms
       JSON.generate    610.348k (± 0.7%) i/s -      3.087M in   5.057383s
             Oj.dump      4.454M (± 0.8%) i/s -     22.611M in   5.077251s
    Oj.dump (compat)    585.849k (± 0.9%) i/s -      2.960M in   5.052141s
     Oj.dump (rails)      1.012M (± 0.6%) i/s -      5.062M in   5.004297s
```

### After
```
       JSON.generate    60.994k i/100ms
             Oj.dump   443.062k i/100ms
    Oj.dump (compat)   102.736k i/100ms
     Oj.dump (rails)   101.957k i/100ms
       JSON.generate    614.990k (± 1.0%) i/s -      3.111M in   5.058600s
             Oj.dump      4.452M (± 0.9%) i/s -     22.596M in   5.075861s
    Oj.dump (compat)      1.038M (± 0.9%) i/s -      5.240M in   5.047304s
     Oj.dump (rails)      1.006M (± 0.8%) i/s -      5.098M in   5.067830s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'
require 'json'

data = {
  created_at: Time.now,
}

Benchmark.ips do |x|
  x.report('JSON.generate') { JSON.generate(data) }
  x.report('Oj.dump (compat)') { Oj.dump(data, mode: :compat) }
  x.report('Oj.dump (rails)') { Oj.dump(data, mode: :rails) }
end
```

### DIFF
--- a/ext/oj/dump_compat.c
+++ b/ext/oj/dump_compat.c
@@ -645,6 +645,16 @@ dump_float(VALUE obj, int depth, Out out, bool as_ok) {
     *out->cur = '\0';
 }
 
+static void
+dump_time(VALUE obj, int depth, Out out, bool as_ok) {
+    if ((as_ok && rb_respond_to(obj, oj_to_json_id)) &&
+        (Yes == out->opts->to_json || rb_respond_to(obj, oj_as_json_id))) {
+        dump_to_json(obj, out);
+        return;
+    }
+    oj_dump_ruby_time(obj, out);
+}
+
 static int
 hash_cb(VALUE key, VALUE value, VALUE ov) {
     Out	out = (Out)ov;
@@ -765,6 +775,11 @@ dump_obj(VALUE obj, int depth, Out out, bool as_ok) {
     if (Yes == out->opts->raw_json && rb_respond_to(obj, oj_raw_json_id)) {
 	oj_dump_raw_json(obj, depth, out);
 	return;
+    }
+
+    if (rb_cTime == rb_obj_class(obj)) {
+        dump_time(obj, depth, out, as_ok);
+        return;
     }
     if (as_ok && rb_respond_to(obj, oj_to_json_id)) {
 	dump_to_json(obj, out);
--- a/ext/oj/dump_compat.c
+++ b/ext/oj/dump_compat.c
@@ -800,12 +800,12 @@ dump_obj(VALUE obj, int depth, Out out, bool as_ok) {
 	return;
     }
 
-    if (as_ok && rb_respond_to(obj, oj_as_json_id)) {
-        dump_as_json(obj, depth, out, true);
-        return;
-    }
     if (rb_cTime == rb_obj_class(obj)) {
-        oj_dump_ruby_time(obj, out);
+        if (as_ok && rb_respond_to(obj, oj_as_json_id)) {
+            dump_as_json(obj, depth, out, true);
+        } else {
+            oj_dump_ruby_time(obj, out);
+        }
         return;
     }
     if (as_ok && rb_respond_to(obj, oj_to_json_id)) {
--- a/ext/oj/dump_compat.c
+++ b/ext/oj/dump_compat.c
@@ -138,6 +138,39 @@ dump_to_json(VALUE obj, Out out) {
     *out->cur = '\0';
 }
 
+static void dump_as_json(VALUE obj, int depth, Out out, bool as_ok) {
+    volatile VALUE ja;
+
+    if (Yes == out->opts->trace) {
+        oj_trace("as_json", obj, __FILE__, __LINE__, depth + 1, TraceRubyIn);
+    }
+    // Some classes elect to not take an options argument so check the arity
+    // of as_json.
+    if (0 == rb_obj_method_arity(obj, oj_as_json_id)) {
+        ja = rb_funcall(obj, oj_as_json_id, 0);
+    } else {
+        ja = rb_funcall2(obj, oj_as_json_id, out->argc, out->argv);
+    }
+    if (Yes == out->opts->trace) {
+        oj_trace("as_json", obj, __FILE__, __LINE__, depth + 1, TraceRubyOut);
+    }
+
+    out->argc = 0;
+    if (ja == obj || !as_ok) {
+        // Once as_json is called it should never be called again on the same
+        // object with as_ok.
+        oj_dump_compat_val(ja, depth, out, false);
+    } else {
+        int type = rb_type(ja);
+
+        if (T_HASH == type || T_ARRAY == type) {
+            oj_dump_compat_val(ja, depth, out, true);
+        } else {
+            oj_dump_compat_val(ja, depth, out, true);
+        }
+    }
+}
+
 static void
 dump_array(VALUE a, int depth, Out out, bool as_ok) {
     size_t	size;
@@ -645,16 +678,6 @@ dump_float(VALUE obj, int depth, Out out, bool as_ok) {
     *out->cur = '\0';
 }
 
-static void
-dump_time(VALUE obj, int depth, Out out, bool as_ok) {
-    if ((as_ok && rb_respond_to(obj, oj_to_json_id)) &&
-        (Yes == out->opts->to_json || rb_respond_to(obj, oj_as_json_id))) {
-        dump_to_json(obj, out);
-        return;
-    }
-    oj_dump_ruby_time(obj, out);
-}
-
 static int
 hash_cb(VALUE key, VALUE value, VALUE ov) {
     Out	out = (Out)ov;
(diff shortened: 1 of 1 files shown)


ext/oj/dump_compat.c
```diff
@@ -645,6 +645,16 @@ dump_float(VALUE obj, int depth, Out out, bool as_ok) {
     *out->cur = '\0';
 }
 
+static void
+dump_time(VALUE obj, int depth, Out out, bool as_ok) {
+    if ((as_ok && rb_respond_to(obj, oj_to_json_id)) &&
+        (Yes == out->opts->to_json || rb_respond_to(obj, oj_as_json_id))) {
```
ohler55, line 651: Since `#to_json` is going to be called, there is no need to check for `#as_json`.  Maybe something like this would be better:

```c
if (as_ok && rb_respond_to(obj, oj_to_json_id)) {
```

ohler55 (changes requested):

ext/oj/dump_compat.c
Watson1978: https://github.com/ohler55/oj/blob/38be7638f2933b123d4e67ac8f40da06617e346b/ext/oj/dump_compat.c#L699

After executed above line, `as_ok` will always have `true`.
Therefore, I add  #as_json check.

Watson1978: And, if it don't call #to_json when #as_json exists, the test in oj will fail....

Watson1978: `#to_json` is so slow, so I don't want to call it if possible.

ohler55: Right, it is slow but it doesn't make sense to check for `#as_json` and then call `#to_json`.

Logic to call `#as_json` instead would be ok but the respond_to check should match the funcall.

Watson1978: If `Time#to_json` is overridden by `require 'json/add/time'`, it need to call `Time#to_json` because behavior is changed.

I want to know a way whether the method is overridden.... but...
So, this PR is checking that `Time#as_json` is added by `require 'json/add/time'` (because I think it can be assumed that `Time#to_json` has been overridden by `require 'json/add/time'`)

ohler55: Is there a more direct way to check? By keying off of `#as_json` any other gem or use code that added `#as_json` would match as well. Is there a way to determine where a method was defined? Maybe checking the method object or the method `#hash`?

ohler55: I wonder if the best approach might be to monkey patch time as part of mimic_JSON. What do you think?

Watson1978: I tried to simplify the branching by calling as_json beforehand.
https://github.com/ohler55/oj/pull/694/commits/127b5a3ae03c7eb896423c823b24b73f647efd43

Watson1978: hmm, fails the unit test...

> I wonder if the best approach might be to monkey patch time as part of mimic_JSON.

I will consider your approach.

ext/oj/dump_compat.c
```diff
@@ -138,6 +138,39 @@ dump_to_json(VALUE obj, Out out) {
...
+
+    out->argc = 0;
+    if (ja == obj || !as_ok) {
+        // Once as_json is called it should never be called again on the same
+        // object with as_ok.
+        oj_dump_compat_val(ja, depth, out, false);
+    } else {
+        int type = rb_type(ja);
+
+        if (T_HASH == type || T_ARRAY == type) {
```
ohler55, line 166: This does seem to do anything as the exact call and argument are used in both cases.

Watson1978:
OK, the specs has been passed.

ext/oj/dump_compat.c
```diff
@@ -766,6 +799,15 @@ dump_obj(VALUE obj, int depth, Out out, bool as_ok) {
...
     }
+
+    if (rb_cTime == rb_obj_class(obj)) {
+        if (as_ok && rb_respond_to(obj, oj_as_json_id)) {
+            dump_as_json(obj, depth, out, true);
+        } else {
+            oj_dump_ruby_time(obj, out);
+        }
+        return;
+    }
```
ohler55, line 810: If my reading of this is correct, `Time.to_json` will never be called. Is that the intent and if so I would expect there will be issues raised it these changes are released,

ohler55:
Two comments to address first. The first is easy. The second may require some re-thinking.

ext/oj/dump_compat.c
Watson1978: Exactly, you're right. To improve performance, like a `rails` mode, I want to avoid Time#to_json calling with this PR.

ohler55: The problem with that approach is that if someone defines a `#to_json` they will expect it to be called. With this change it will not be called yet with the JSON it would so Oj in compat mode would not longer be compatible with the JSON gem.

Watson1978: Yeah, I have same concerns from the beginning, so this patch e focus on Time class only.

Watson1978: I think anyone does not override Time#to_json method.

ohler55: So maybe a check that the dump is in rails mode is needed or possibly just make sure the rails dump code is patched by not the compat code.

Watson1978: If we use `method` and `super_method`, we may be able to know if `to_json` was overridden.

Here is a sample code.

```ruby
class Foo
  def say(arg)
    puts arg
  end
end

class Bar < Foo
  def say(arg)
    super
  end
end

p Bar.new.method(:say).super_method # => #<Method: Foo#say>
p Foo.new.method(:say).super_method # => nil
```

ohler55: If the goal is to just improve the Rails handling of time then I think it would be better to look at the rails.c file. In particular, the `dump_obj()` function near the bottom of the file. If in rails mode all the type including time should have an `#as_json` method which should be called in preference to the `#to_json`. 

As for not overriding the `#to_json` method on time it is hard to know what people will do. It has happened more than once that I changed or removed something I though was never used only to have an issue created by some use that was using what I expected could be changed. We don't have a very good feedback cycle with gem feature use I'm afraid.

Watson1978: This patch already calls #as_json in preference to #to_json...

hmm, am I mistaken about something?

ohler55: The patch checks for `#as_json` but if that is not present it drops back to the calling `oj_dump_ruby_time()` which just calls `#to_s` on Time. No chance to run the `#to_json` so it is not that there is a preference for `#as_json` over `#to_json` but rather it is either `#as_json` or nothing (well `#to_s`).

ohler55: Maybe it would help if the use case were more clear. I think you are trying to solve the issue for Rails but not for direct JSON gem use, right?

ohler55: If I am missing something I would be glad to get on a call with you. Zoom of something.

Watson1978: The main motivation:

I hope to use only Oj gem in my rails app. However, the gems which exists somewhere in dependency requires JSON gem. And it decreases a Time#to_json performance. I want to solve my problem

ohler55: I see. It sounds like the best option would be to see if the `:rails` mode is in use. If it is in your tests then it would be okay to not check for `#to_json` I think. Can you check? That might be the way to get around the slow `#to_json`.

Watson1978: The JSON API in my rails app uses oj with :compat mode ... hmm.

ohler55: That's a tough one. That implies the author wanted JSON gem behaviour of calling `#to_json`. That or didn't know any better. Is it your code or can it be monkey patched?

Watson1978: It seems difficult to merge with my little spare time, so close it.

ohler55: Sorry to hear that.

Watson1978: Sorry to spend your time and I hope that you have an overview of this problem.

I'll try again when I have more time :)

ohler55:
Not a waste of time at all. It was nice working with you.

PR #695, opened 2021-08-21
Reduce memory usage in `Oj.load` with using Ruby 3.0 feature

Watson1978:
`rb_enc_interned_str()` was added in Ruby 3.0 which returns frozen string object.

It has same effect with `# frozen_string_literal: true` magic comment and
The object will be reused and reduce memory usage if you create a frozen String
with the same C string over and over.

Here is memory usage with sample code.

−               | before   | after
--               | --       | --
Oj.dump          | 33.904k  | 25.904k

### Environment
- MacBook Air (M1, 2020)
- macOS 12.0 beta 4
- Apple M1
- Ruby 3.0.2

### Before
```
[CPU]====================
             Oj.load     3.855k i/100ms
             Oj.load     38.190k (± 0.7%) i/s -    192.750k in   5.047359s
[MEMORY]====================
             Oj.load    33.904k memsize (     0.000  retained)
                       502.000  objects (     0.000  retained)
                         4.000  strings (     0.000  retained)
```

### After
```
[CPU]====================
             Oj.load     3.758k i/100ms
             Oj.load     37.368k (± 0.9%) i/s -    187.900k in   5.028814s
[MEMORY]====================
             Oj.load    25.904k memsize (     0.000  retained)
                       302.000  objects (     0.000  retained)
                         2.000  strings (     0.000  retained)
```

### Test code
Before, following objects will be generated for one Hash object

- A Hash object
- Two String as Hash key
- Two String as Hash value

After, the same hash keys will be reused.
(The following objects will be generated for one Hash object with this patch)

- A Hash object
- Two String as Hash value

```ruby
require 'benchmark/ips'
require 'benchmark/memory'
require 'oj'

item = {
  "type" => "string",
  "description" => "The important things"
}

data = (1..100).map { item }
json = Oj.dump(data)

puts "[CPU]" + "=" * 20
Benchmark.ips do |x|
  x.report('Oj.load') { Oj.load(json) }
end

puts "[MEMORY]" + "=" * 20
Benchmark.memory do |x|
  x.report('Oj.load') { Oj.load(json) }
end
```

### DIFF
--- a/ext/oj/strict.c
+++ b/ext/oj/strict.c
@@ -35,18 +35,22 @@ VALUE oj_calc_hash_key(ParseInfo pi, Val parent) {
         if (Yes == pi->options.sym_key) {
             rkey = ID2SYM(rb_intern3(parent->key, parent->klen, oj_utf8_encoding));
         } else {
+#if HAVE_RB_ENC_INTERNED_STR
+            rkey = rb_enc_interned_str(parent->key, parent->klen, oj_utf8_encoding);
+#else
             rkey = rb_str_new(parent->key, parent->klen);
             rkey = oj_encode(rkey);
+            OBJ_FREEZE(rkey);
+#endif
         }
-        OBJ_FREEZE(rkey);
         return rkey;
     }
     if (Yes == pi->options.sym_key) {
 	rkey = oj_sym_intern(parent->key, parent->klen);
     } else {
 	rkey = oj_str_intern(parent->key, parent->klen);
-    }
     OBJ_FREEZE(rkey);
+    }
     return rkey;
 }
 
--- a/ext/oj/extconf.rb
+++ b/ext/oj/extconf.rb
@@ -31,6 +31,7 @@ have_func('rb_gc_mark_movable')
 have_func('stpcpy')
 have_func('pthread_mutex_init')
 have_func('rb_enc_associate')
+have_func('rb_enc_interned_str')
 have_func('rb_ext_ractor_safe', 'ruby.h')
 # rb_hash_bulk_insert is deep down in a header not included in normal build and that seems to fool have_func.
 have_func('rb_hash_bulk_insert', 'ruby.h') unless '2' == version[0] && '6' == version[1]
--- a/ext/oj/object.c
+++ b/ext/oj/object.c
@@ -64,12 +64,16 @@ static VALUE calc_hash_key(ParseInfo pi, Val kval, char k1) {
         if (Yes == pi->options.sym_key) {
             rkey = ID2SYM(rb_intern3(kval->key, kval->klen, oj_utf8_encoding));
         } else {
+#if HAVE_RB_ENC_INTERNED_STR
+            rkey = rb_enc_interned_str(kval->key, kval->klen, oj_utf8_encoding);
+#else
             rkey = rb_str_new(kval->key, kval->klen);
             rkey = oj_encode(rkey);
+            OBJ_FREEZE(rkey);
+#endif
         }
     }
 #endif
-    OBJ_FREEZE(rkey);
     return rkey;
 }
 
--- a/ext/oj/wab.c
+++ b/ext/oj/wab.c
@@ -305,11 +305,15 @@ static VALUE calc_hash_key(ParseInfo pi, Val parent) {
     if (Yes == pi->options.cache_keys) {
         rkey = oj_sym_intern(parent->key, parent->klen);
     } else {
+#if HAVE_RB_ENC_INTERNED_STR
+        rkey = rb_enc_interned_str(parent->key, parent->klen, oj_utf8_encoding);
+#else
         rkey = rb_str_new(parent->key, parent->klen);
         rkey = oj_encode(rkey);
         rkey = rb_str_intern(rkey);
+        OBJ_FREEZE(rkey);
+#endif
     }
-    OBJ_FREEZE(rkey);
     return rkey;
 }
 


ohler55:
Let's discuss this a bit. I did some benchmarks when I added the cache option and my cache ended up being faster so when someone select cached keys that cache is used. The reason it is optional is that when parsing something like a dictionary with 100,000 keys the caching locks up a large amount of memory that is is not collected during a GC. Calling the Ruby intern functions does the same thing I think. If it does then it is best not to intern or cache when the options is turned off. Can you verify that there is code in Ruby to expire the interned entries? I would expect not and if not then the Ruby intern functions should not be called when the caching option is off. Would you agree or have I missed something?

ohler55:
You did give me an idea though. Oj make no claim that the keys or strings are interned forever but rather just as an optimization. Maybe I could set the cache to have a limited number of entries and then flush out less used entries to save space.

Watson1978:
At least, looks for me that it checks wheather the frozen string has been garbaged when we will register the string in `fstr_update_callback()`
- https://github.com/ruby/ruby/blob/18031f4102d7c8efc3a40b3fa65b36f94fc8a418/string.c#L388-L408
- https://github.com/ruby/ruby/blob/18031f4102d7c8efc3a40b3fa65b36f94fc8a418/string.c#L299-L348

It is also used in several libraries and the cache will be used effectively across these libraries.

- json
  - https://github.com/flori/json/blob/5ed8cf7eb1e60eba81909336a540437fd74745d3/ext/json/ext/parser/parser.c#L2457
- msgpack-ruby
  - https://github.com/msgpack/msgpack-ruby/blob/77e75afe33c3908497c98e8916845527237bc29c/ext/msgpack/buffer.h#L459
- mysql2
  - https://github.com/brianmario/mysql2/blob/7f4e844fccf6afa888d0bd108d4707a2a7784484/ext/mysql2/result.c#L175

ohler55:
I agree with your point that a shared intern cache is preferred. I'm not sure that is justification for either dropping the faster Oj cache or for interning all key strings. Even in the example you pointed out, the json gem only interns if the "intern" flag is true.

The `faster_update_callback` does show that frozen strings are collected at some point but not interned strings.

Stepping back lets look at what is desired. We want faster string creation or reuse on strings that are encountered frequently. Unless we can assure that the cache does lock up too much memory there should be some way of making sure that does not happen. Right now there is an option to turn on or off the caching of keys as well as shorter strings. That solution is not ideal but for now lets assume it should be use until we find a better way should as a auto-expiring cache. If the cache option is to be honoured then the Ruby intern functions should only be called when the cache option is true. That leaves us with which is better, the Oj cache or the Ruby intern. When I benchmarked the two options Oj gave better results. In favour of the Ruby intern though is that the intern cache is shared across all gems so depending on the gems in use the memory use may be lower. The choice then is which is preferred better performance or potential reduced memory use?

I have been giving some thought to ache control options other than a flag. Some possibilities are:
 * an auto-expiring cache based on most recently used
 * an option to flush the cache on GC
 * flush cache entries not used since the last GC
 * provide a call like `Oj.clear_cache`

Any of those options would allow the cache to always be on as the issue of unbounded memory use from huge dictionary files such as the one user tried that had over 100,000 keys in a Hash.

Watson1978:
The objects cached by rb_enc_interned_str() are also garbaged with GC

Here is an easy experiment.

```diff
diff --git a/ext/oj/oj.c b/ext/oj/oj.c
index f01f656..d995db2 100644
--- a/ext/oj/oj.c
+++ b/ext/oj/oj.c
@@ -1748,6 +1748,13 @@ static VALUE protect_require(VALUE x) {
     return Qnil;
 }
 
+static VALUE oj_fstring(VALUE self, VALUE str) {
+#if HAVE_RB_ENC_INTERNED_STR
+    return rb_enc_interned_str(RSTRING_PTR(str), RSTRING_LEN(str), rb_utf8_encoding());
+#endif
+    rb_raise(rb_eRuntimeError, "Non supported Ruby version.");
+}
+
 /* Document-module: Oj
  *
  * Optimized JSON (Oj), as the name implies was written to provide speed
@@ -1809,6 +1816,8 @@ void Init_oj() {
     rb_define_module_function(Oj, "object_load", oj_object_parse, -1);
     rb_define_module_function(Oj, "wab_load", oj_wab_parse, -1);
 
+    rb_define_module_function(Oj, "fstring", oj_fstring, 1);
+
     rb_define_module_function(Oj, "dump", dump, -1);
 
     rb_define_module_function(Oj, "to_file", to_file, -1);
```

```ruby
require 'oj'

1_000_000.times do |i|
  Oj.fstring "key_#{i}"

  if i % 10000 == 0
    puts "#{i},#{rss}"
  end
end
```

## Result
```
$ ruby memory.rb
0,14.73828125
10000,15.39453125
20000,15.62109375
30000,15.8515625
40000,15.890625
50000,15.90234375
60000,15.921875
70000,15.921875
80000,15.921875
90000,15.921875
100000,15.921875
110000,15.921875
120000,15.921875
130000,15.921875
140000,15.921875
150000,15.921875
160000,15.921875
170000,15.921875
180000,15.9296875
190000,15.9296875
200000,15.9296875
210000,15.9375
220000,15.9375
230000,15.9375
240000,15.9375
250000,15.9375
260000,15.9375
270000,15.9375
280000,15.9375
290000,15.9375
300000,15.9375
310000,15.9375
320000,15.9375
330000,15.9375
340000,15.9375
350000,15.9375
360000,15.9375
370000,15.9375
380000,15.9375
390000,15.9375
400000,15.9375
410000,15.9375
420000,15.9375
430000,15.9375
440000,15.9375
450000,15.9375
460000,15.9375
470000,15.9375
480000,15.9375
490000,15.9375
500000,15.9375
510000,15.9375
520000,15.9375
530000,15.9375
540000,15.9375
550000,15.9375
560000,15.9375
570000,15.9375
580000,15.9375
590000,15.9375
600000,15.9375
610000,15.9375
620000,15.9375
630000,15.9375
640000,15.9375
650000,15.9375
660000,15.9375
670000,15.9375
680000,15.9375
690000,15.9375
700000,15.9375
710000,15.9375
720000,15.9375
730000,15.9375
740000,15.9375
750000,15.9375
760000,15.9375
770000,15.9375
780000,15.9375
790000,15.9375
800000,15.9375
810000,15.9375
820000,15.94921875
830000,15.94921875
840000,15.94921875
850000,15.94921875
860000,15.94921875
870000,15.94921875
880000,15.94921875
890000,15.94921875
900000,15.94921875
910000,15.94921875
920000,15.94921875
930000,15.94921875
940000,15.94921875
950000,15.94921875
960000,15.94921875
970000,15.94921875
980000,15.94921875
990000,15.94921875
```

If the object created by rb_enc_interned_str() is persistent, it must be using more huge amount of memory.

ohler55:
That’s great! It also means I need to review the code to make sure there are no assumptions that an interned string will not be GCed. I’ll review and approve later tonight, EST.

ext/oj/strict.c
```diff
@@ -35,18 +35,22 @@ VALUE oj_calc_hash_key(ParseInfo pi, Val parent) {
...
+#endif
         }
-        OBJ_FREEZE(rkey);
         return rkey;
     }
     if (Yes == pi->options.sym_key) {
 	rkey = oj_sym_intern(parent->key, parent->klen);
     } else {
 	rkey = oj_str_intern(parent->key, parent->klen);
-    }
```
ohler55, line 48: Line 48 should not be removed and line 543 should not be added. Both `oj_str_intern` already  freezes the strings. As for `oj_sym_intern`, it does not only because my understanding is that symbols are immutable. If that is not the case then a freeze is needed in the oj intern.c and not here.

ohler55: My error. Misread the change. Good catch.

ohler55:
Some bad news. I'm not going to be able to include the change in the next release. I added a test in test/perf_once.rb. It is a simplified version of a test a user provided for loading a file with a Hash that has just under 750,000 keys. It is an example of when caching should not be used. With the Ruby intern function used the loading takes 50% longer. The Ruby intern use is faster than the Oj cache in this situation by a significant amount which is different than the results with the test/perf_parser.rb results.

It looks like there might be a few improvements to make to the Oj cache. One is auto expiration of entries and the other is handling more than 10,000 entries more efficiently.

SamSaffron:
@ohler55 @Watson1978 I think there is a possibility this change is causing memory issues for us.

We are seeing segfaults and stuck processes. 

I just ran a gdb on one of the stuck processes and am seeing: 

```
[Switching to thread 38 (Thread 0x7fe38f9fe700 (LWP 32643))]
#0  0x00007fe3d28a2920 in locking_intern (c=0x7fe3d4f8a710, 
    key=0x7fe37e2b770c "__tlds_replaced__\":true,\"re\":{\"src_Any\":\"[\\\\0-\\\\uD7FF\\\\uE000-\\\\uFFFF]|[\\\\uD800-\\\\uDBFF][\\\\uDC00-\\\\uDFFF]|[\\\\uD800-\\\\uDBFF](?![\\\\uDC00-\\\\uDFFF])|(?:[^\\\\uD800-\\\\uDBFF]|^)[\\\\uDC00-\\\\uDFFF]\",\"src_Cc\":\"[\\\\"..., len=17) at cache.c:198
198	cache.c: No such file or directory.
(gdb) bt
#0  0x00007fe3d28a2920 in locking_intern (c=0x7fe3d4f8a710, 
    key=0x7fe37e2b770c "__tlds_replaced__\":true,\"re\":{\"src_Any\":\"[\\\\0-\\\\uD7FF\\\\uE000-\\\\uFFFF]|[\\\\uD800-\\\\uDBFF][\\\\uDC00-\\\\uDFFF]|[\\\\uD800-\\\\uDBFF](?![\\\\uDC00-\\\\uDFFF])|(?:[^\\\\uD800-\\\\uDBFF]|^)[\\\\uDC00-\\\\uDFFF]\",\"src_Cc\":\"[\\\\"..., len=17) at cache.c:198
#1  0x00007fe3d28da0aa in oj_calc_hash_key (pi=<optimized out>, parent=<optimized out>) at strict.c:47
#2  0x00007fe3d28a413d in hash_set_value (pi=0x7fe38f8f91e0, parent=0x7fe38f8fa540, value=20) at compat.c:158
#3  0x00007fe3d28c38a7 in add_value (pi=0x7fe38f8f91e0, rval=<optimized out>) at parse.c:83
#4  0x00007fe3d28c6240 in read_true (pi=0x7fe38f8f91e0) at parse.c:119

``` 

Very hard to tell if this is exactly the culprit but it appears this area changed. We are collecting core dumps and are on `3.13.3`

ohler55:
@SamSaffron would you mind opening an issue for that? It does look like the new cache is the issue based on the stack trace. Can you describe a bit more about what the setting were and what the application was? Were multiple threads used?

SamSaffron:
sure, opening it now!

PR #708, opened 2021-09-25
Fix `Oj::Doc` behaviour for inexisting path

### LINKED ISSUE #707: `Oj::Doc#fetch` does not use default value when part of the part exists
Reproduction:

```ruby
json = '{"foo":null}'
doc=Oj::Doc.open(json)

puts "Should all be true:"
p doc.fetch("/") == {"foo"=>nil}
p doc.fetch("/fizz/bar", :default_should_be_taken) == :default_should_be_taken
p doc.fetch("/foo/bar", :default_should_be_taken) == :default_should_be_taken # failure
```

I _may_ have some time to work on that one later (at least in a month)..

It seems that the issue comes from `get_doc_leaf` since `Oj::Doc#exists?` has the same bug

BuonOmo:
```ruby
Oj::Doc.open('{"foo":1}') do |doc|
  doc.fetch('/foo/bar') # used to give `1`, now gives `nil`
  doc.exists?('/foo/bar') # used to give `true`, now gives `false`
end
```

Closes #707

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+- Fix `Oj::Doc` behaviour for inexisting path.
+  ```ruby
+  Oj::Doc.open('{"foo":1}') do |doc|
+    doc.fetch('/foo/bar') # used to give `1`, now gives `nil`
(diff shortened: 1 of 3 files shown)


ohler55:
Thanks for the fix.

PR #709, opened 2021-09-27
Remove the code for Ruby 1.8

Watson1978:
Remove code for unsupported Ruby versions.

Ref:
https://github.com/ohler55/oj/blob/a15eb1a305c34574199cfa33c3aae0b9fe31791e/oj.gemspec#L23


ohler55:
Oj has been around long enough to have collected a few of those left over bits of code. Thanks for the cleanup.

Issue #693, opened 2021-08-12
Strange behavior serializing BigDecimal with JSON.generate


mrbongiolo:
Hi @ohler55, thanks for your amazing work on Oj! I found an "issue" in the way `BigDecimal` is handled when upgrading from v3.11.3 to v3.11.4 when using `JSON.generate`.

It's a rails project with using those configs:
```ruby
Oj.optimize_rails
Oj.default_options = {
  mode: :rails,
  float_precision: 0,
  bigdecimal_as_decimal: true,
  bigdecimal_load: :bigdecimal,
  trace: true
}
```

When using v3.11.3:
```
[1] pry(main)> JSON.generate({:ticker=>"ASAI3", :price=>18.7.to_d, :rate=>-0.8.to_d})
#0:dump_compat.c:938:Oj:}: dump Hash
#0:dump_compat.c:938:Oj:}:   dump String
#0:dump_compat.c:963:Oj:{:   dump String
#0:dump_compat.c:938:Oj:}:   dump BigDecimal
#0:dump_compat.c:120:Oj:>: to_json BigDecimal
#0:      rails.c:1493:Oj:}: dump BigDecimal
#0:      rails.c:1504:Oj:{: dump BigDecimal
#0:dump_compat.c:128:Oj:<: to_json BigDecimal
#0:dump_compat.c:963:Oj:{:   dump BigDecimal
#0:dump_compat.c:938:Oj:}:   dump BigDecimal
#0:dump_compat.c:120:Oj:>: to_json BigDecimal
#0:      rails.c:1493:Oj:}: dump BigDecimal
#0:      rails.c:1504:Oj:{: dump BigDecimal
#0:dump_compat.c:128:Oj:<: to_json BigDecimal
#0:dump_compat.c:963:Oj:{:   dump BigDecimal
#0:dump_compat.c:963:Oj:{: dump Hash
=> "{\"ticker\":\"ASAI3\",\"price\":18.7,\"rate\":-0.8}"

[3] pry(main)> {:ticker=>"ASAI3", :price=>18.7.to_d, :rate=>-0.8.to_d}.to_json
#0:      rails.c:1493:Oj:}: dump Hash
#0:      rails.c:1493:Oj:}:   dump String
#0:      rails.c:1504:Oj:{:   dump String
#0:      rails.c:1493:Oj:}:   dump BigDecimal
#0:      rails.c:1504:Oj:{:   dump BigDecimal
[...]

PR #716, opened 2021-10-14
Update usage of :bigdecimal_as_decimal in Rails Mode

mrbongiolo:
The option `:bigdecimal_as_decimal` is not available in Rails Mode (as discussed here https://github.com/ohler55/oj/issues/693), as the original JSON and Rails implementation does not have a way to "force" bigdecimal values to be encoded as numbers in json.


ohler55:
It took a while but I did finally get a break from moving to look at this. There is a way to set the `:bigdecimal_as_decimal` with rails. If after setting everything else up the default options are set for the `:bigdecimal_as_decimal` it will be honoured by the rails mode. It can not be provided as an argument but can be set with the default options.

PR #719, opened 2021-11-05
Use rb_type() to check primitive class type

Watson1978:
It seems slightly faster to use `rb_type()` to check for primitive class types than `rb_obj_class()​`.

Oj.dump          | 439.927k | 442.567k | 1.006x

### Environment
- MacBook Pro (M1 Max, 2021)
- macOS 12.0
- Apple M1 Max
- Ruby 3.0.2

### Before
```
Oj.load(symbol_keys)    44.036k i/100ms
Oj.load(symbol_keys)    439.927k (± 0.3%) i/s -      4.404M in  10.009941s
```

### After
```
Oj.load(symbol_keys)    44.146k i/100ms
Oj.load(symbol_keys)    442.567k (± 0.3%) i/s -      4.459M in  10.074843s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

json =<<-EOF
{
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
    },
    "lastName": {
      "type": "string",
    },
    "age": {
      "type": "integer",
      "minimum": 0
    }
  }
}
EOF

Benchmark.ips do |x|
  x.warmup = 10
  x.time = 10

  x.report('Oj.load(symbol_keys)') { Oj.load(json, symbol_keys:true) }
end
```

### DIFF
--- a/ext/oj/object.c
+++ b/ext/oj/object.c
@@ -536,7 +536,7 @@ WHICH_TYPE:
         }
         break;
     case T_HASH:
-        if (rb_cHash != rb_obj_class(parent->val)) {
+        if (T_HASH != rb_type(parent->val)) {
             if (4 == klen && 's' == *key && 'e' == key[1] && 'l' == key[2] && 'f' == key[3]) {
                 rb_funcall(parent->val, oj_replace_id, 1, value);
             } else {
--- a/ext/oj/compat.c
+++ b/ext/oj/compat.c
@@ -45,7 +45,7 @@ static void hash_set_cstr(ParseInfo pi, Val kval, const char *str, size_t len, c
                 rstr = rb_funcall(clas, oj_json_create_id, 1, rstr);
             }
         }
-        if (rb_cHash != rb_obj_class(parent->val)) {
+        if (T_HASH != rb_type(parent->val)) {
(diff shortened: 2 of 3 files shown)


ext/oj/object.c
```diff
@@ -536,7 +536,7 @@ static void hash_set_value(ParseInfo pi, Val kval, VALUE value) {
         }
         break;
     case T_HASH:
-        if (rb_cHash != rb_obj_class(parent->val)) {
+        if (T_HASH != rb_type(parent->val)) {
```
Watson1978, line 539: This is already checked in the previous switch statement.
But I didn't understand why this line is checking it again...

ohler55: In some cases the Hash has been subclassed with the expectation that the set method will have a side effect. Checking the class explicitly catches that.

ohler55:
It looks like the changes fail due to the difference between an object being a Hash type versus being of the Hash class.

ext/oj/object.c
Watson1978: I see.
Hmm, this PR breaks the operation with subclasses.
So, I will close this PR.

Thanks

PR #724, opened 2021-11-27
Use RARRAY_AREF() instead to improve array accessing

Watson1978:
`rb_ary_entry()` includes boundary value handling, etc.

```c
static inline VALUE
rb_ary_entry_internal(VALUE ary, long offset)
{
    long len = RARRAY_LEN(ary);
    const VALUE *ptr = RARRAY_CONST_PTR_TRANSIENT(ary);
    if (len == 0) return Qnil;
    if (offset < 0) {
        offset += len;
        if (offset < 0) return Qnil;
    }
    else if (len <= offset) {
        return Qnil;
    }
    return ptr[offset];
}
```

Oj has simply array handling and it does not need them.
This patch will cut off the these overheads.

−               | before | after  | result
--               | --     | --     | --
Oj.dump          | 1.593M | 1.687M | 1.059x

### Environment
- MacBook Pro (14 inch, 2021)
- macOS 12.0
- Apple M1 Max
- Ruby 3.0.2

### Before
```
             Oj.dump   159.458k i/100ms
             Oj.dump      1.593M (± 0.6%) i/s -     15.946M in  10.010658s
```

### After
```
             Oj.dump   168.482k i/100ms
             Oj.dump      1.687M (± 0.4%) i/s -     17.017M in  10.084360s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

data = {
  "strings": ("a".."z").to_a,
}

Benchmark.ips do |x|
  x.warmup = 10
  x.time = 10

end
```

### DIFF
--- a/ext/oj/custom.c
+++ b/ext/oj/custom.c
@@ -749,7 +749,7 @@ static void dump_array(VALUE a, int depth, Out out, bool as_ok) {
             } else {
                 fill_indent(out, d2);
             }
-            oj_dump_custom_val(rb_ary_entry(a, i), d2, out, true);
+            oj_dump_custom_val(RARRAY_AREF(a, i), d2, out, true);
             if (i < cnt) {
                 *out->cur++ = ',';
             }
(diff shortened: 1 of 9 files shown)


ohler55:
Thank you!

PR #727, opened 2021-12-16
Actions - Add Windows CI

MSP-Greg:
This adds basic CI for Windows, Ruby 2.6 thru head (ucrt)

1. '[CI] - add windows-2022 2.6 - master' - workflow - added a separate job for Windows.  Also, removed the restriction on branches, which allows people to run CI in their forks.  If you do work here in another branch and the merge the branches, this may run CI twice.  I can revert.

2. 'test_saj.rb - fix for windows paths' - simple fix for windows paths

3. '[CI] rakefile & .gitignore' - Rakefile - Make cmd handling cross-platform.  Actions CI does 'odd' things with STDOUT, especially on Windows.  Added changes in rakefile are to display output in correct order.  Added tmp dir to .gitignore.

Happy to change anything...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,8 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - develop
-      - master
+      - '*'
   pull_request:
 
 jobs:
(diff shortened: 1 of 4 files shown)


ohler55:
Awesome! Thank you.

MSP-Greg:
You're welcome, glad I could help.  I thought CI for compiling and the basic tests was a good start, not sure how important the other tests are?

There were a couple of compile warnings, I'm not much for cross-platform c problems.

Lastly, in the past, Windows Rubies were either mingw or mswin, with very few people using mswin.  The only publicly available builds were mingw.  The mingw tools are part of MSYS2, and they recently released another set of tools known as 'ucrt'.  'mingw' linked to a legacy (but probably supported by MSFT for old enterprise software) runtime often referred to as msvcrt.  The new 'ucrt' tools are linked to the same runtime that all newer windows software is built with.  Window Ruby 3.1 will probably only be released as a 'ucrt' build...

ohler55:
That's all good to know. I'm not a Windows user so having someone who can keep the Windows aspect working is very helpful.

PR #728, opened 2021-12-18
Unroll the loop of string size calculation

Watson1978:
Unrolling simple processing loops seems to improve performance when compiled with GCC.

−               | before | after  | result
--               | --     | --     | --
Oj.dump (macOS)  | 1.016M | 1.048M | 1.031x
Oj.dump (Linux)  | 0.984M | 1.143M | 1.162x

### Environment
- macOS
  - macOS 12.1
  - Apple M1 Max
  - Apple clang version 13.0.0 (clang-1300.0.29.30)
  - Ruby 3.0.2
- Linux
  - Zorin OS 16
  - AMD Ryzen 7 5700G
  - gcc version 9.3.0
  - Ruby 3.0.2

#### Before
```
Warming up --------------------------------------
             Oj.dump   101.770k i/100ms
Calculating -------------------------------------
             Oj.dump      1.016M (± 0.5%) i/s -     10.177M in  10.012877s
```

#### After
```
Warming up --------------------------------------
             Oj.dump   104.842k i/100ms
Calculating -------------------------------------
             Oj.dump      1.048M (± 0.6%) i/s -     10.484M in  10.008101s
```

#### Before
```
Warming up --------------------------------------
             Oj.dump    98.852k i/100ms
Calculating -------------------------------------
             Oj.dump    983.725k (± 1.9%) i/s -      9.885M in  10.052621s
```

#### After
```
Warming up --------------------------------------
             Oj.dump   111.510k i/100ms
Calculating -------------------------------------
             Oj.dump      1.143M (± 1.4%) i/s -     11.486M in  10.050827s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

json =<<-EOF
{
  "$id": "https://example.com/person.schema.json",
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
      "description": "The person's first name."
    },
    "lastName": {
      "type": "string",
      "description": "The person's last name."
    },
    "age": {
      "description": "Age in years which must be equal to or greater than zero.",
      "type": "integer",
      "minimum": 0
    }
  }
}
EOF

Benchmark.ips do |x|
  x.warmup = 10
  x.time = 10

  data = Oj.load(json)
  x.report('Oj.dump') { Oj.dump(data, mode: :compat) }
end
```

PR #729, opened 2021-12-18
Fix string size calculation

Watson1978:
Related to https://github.com/ohler55/oj/pull/728

Previous PR, I had cleaned up at the end of creating it, but I forgot to edit a part.
Sorry.


ohler55:
I guess I should look more carefully. I should have caught that. Thanks.

PR #731, opened 2021-12-30
Test against Rails 7

enomotodev:


ohler55:
Do you want to try and get Ruby 3.2 (ruby head) working? I'm okay with it failing since 3.2 isn't released yet and the failures are not related to your change.

gemfiles/rails_7.gemfile
```diff
@@ -0,0 +1,6 @@
+source "https://rubygems.org"
+
+gem "rails", github: "rails/rails", branch: "7-0-stable"
```
enomotodev, line 3: Rails 7.0.0 is not compatible with Ruby 3.1.0, so need to use Rails `7-0-stable` branch until Rails 7.0.1 is released.

PR #735, opened 2022-01-10
Use snprintf() of standard C library

Watson1978:
Ruby has its own implementation of `ruby_snprintf()`, which is used when `snprintf()` is called.

```c
#define snprintf ruby_snprintf
```
https://github.com/ruby/ruby/blob/cb4e2cb55a59833fc4f1f6db2f3082d1ffcafc80/include/ruby/subst.h#L13

I'm guessing that `ruby_snprintf()` will behave differently from `snprintf()`.

(Ref. https://github.com/ohler55/oj/commit/5e9c70775a9ebdd7262d7c47a70498ad48275c33)

−               | before | after  | result
--               | --     | --     | --
Oj.dump          | 1.459M | 1.894M | 1.298x

### Environment
- MacBook Pro (14 inch, 2021)
- macOS 12.0
- Apple M1 Max
- Ruby 3.1.0

### Before
```
Warming up --------------------------------------
             Oj.dump   145.852k i/100ms
Calculating -------------------------------------
             Oj.dump      1.459M (± 0.8%) i/s -     22.024M in  15.097734s
```

### After
```
Warming up --------------------------------------
             Oj.dump   189.609k i/100ms
Calculating -------------------------------------
             Oj.dump      1.894M (± 0.6%) i/s -     28.441M in  15.020956s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

data = {
  float: 3.141592653589793,
  fixnum: 2 ** 60
}

Benchmark.ips do |x|
  x.time = 15

  x.report('Oj.dump') { Oj.dump(data) }
end
```

PR #742, opened 2022-01-15
Use memcpy() of standard C library

Watson1978:
Ruby has its own implementation of ruby_nonempty_memcpy(), which is used when memcpy() is called.

```c
static inline void *
ruby_nonempty_memcpy(void *dest, const void *src, size_t n)
{
    if (n) {
        return memcpy(dest, src, n);
    }
    else {
        return dest;
    }
}
RBIMPL_SYMBOL_EXPORT_END()
#undef memcpy
#define memcpy ruby_nonempty_memcpy
```
https://github.com/ruby/ruby/blob/master/include/ruby/internal/memory.h

It has an unnecessary `if` statement and it has some overhead.

Similar: https://github.com/ohler55/oj/pull/735

−               | before | after   | result
--               | --     | --      | --
Oj.dump (macOS)  | 7.839k |  8.000k | 1.021x
Oj.dump (Linux)  | 9.140k | 10.043k | 1.099x

### Environment
- macOS
  - macOS 12.1
  - Apple M1 Max
  - Apple clang version 13.0.0 (clang-1300.0.29.30)
  - Ruby 3.1.0
- Linux
  - Zorin OS 16
  - AMD Ryzen 7 5700G
  - gcc version 11.1.0
  - Ruby 3.1.0

#### Before
```
             Oj.dump   783.000  i/100ms
             Oj.dump      7.839k (± 0.9%) i/s -    118.233k in  15.084888s
```

#### After
```
             Oj.dump   803.000  i/100ms
             Oj.dump      8.000k (± 0.8%) i/s -    120.450k in  15.057950s
```

#### Before
```
             Oj.dump   887.000  i/100ms
             Oj.dump      9.140k (± 0.9%) i/s -    137.485k in  15.043569s
```

#### After
```
             Oj.dump   972.000  i/100ms
             Oj.dump     10.043k (± 0.9%) i/s -    150.660k in  15.002236s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

data = (0..10000).to_a

Benchmark.ips do |x|
  x.time = 15

  x.report('Oj.dump') { Oj.dump(data, mode: :compat) }
end
```

### DIFF
--- a/ext/oj/oj.h
+++ b/ext/oj/oj.h
@@ -45,6 +45,10 @@ enum st_retval { ST_CONTINUE = 0, ST_STOP = 1, ST_DELETE = 2, ST_CHECK };
     #include <stdio.h>
 #endif
 
+// To avoid using ruby_nonempty_memcpy().
+#undef memcpy
+#include <string.h>
+
 typedef enum { Yes = 'y', No = 'n', NotSet = 0 } YesNo;
 
 typedef enum {


Watson1978:
Here is stacktrace with before changing.

スクリーンショット 2022-01-16 0 18 06

PR #748, opened 2022-02-15
CI.yml - back to ruby/setup-ruby

MSP-Greg:
ruby/setup-ruby is the repo with the custom action that loads Ruby and sets up the environment for it.

When PR #727 was done, several updates to ruby/setup-ruby were not yet merged. MSP-Greg/ruby-setup-ruby contained many of those updates, so CI was switched to that.

ruby/setup-ruby is now working with Windows-2022 image, so switch back to it.


ohler55:
Thanks

PR #757, opened 2022-04-29
Fix buffer allocation/grow problem and refactor buffer code.

rudihorn:
Currently there are quite a few places where a stack allocated buffer of `N + 10` bytes are allocated instead of `N + 64`. This can become problematic when the buffer is grown and `N + 64` bytes are read, causing the application to read past the actual buffer bounds.

As there are quite a few places where these allocations occur, I've tried to abstract over them by creating the functions `oj_out_init_*` for various stack and heap allocations, as well as an `oj_out_free` function for performing the cleanup. I have placed them in dump.c but there may be a more preferable place to put them. I also did some other small refactoring where I saw that existing functions can be reused.

I am submitting this patch on behalf of Oracle.

### DIFF
--- a/ext/oj/dump.c
+++ b/ext/oj/dump.c
@@ -560,10 +560,7 @@ void oj_dump_obj_to_json(VALUE obj, Options copts, Out out) {
 
 void oj_dump_obj_to_json_using_params(VALUE obj, Options copts, Out out, int argc, VALUE *argv) {
     if (0 == out->buf) {
-        out->buf = ALLOC_N(char, 4096);
-        // 1 less than end plus extra for possible errors
-        out->end       = out->buf + 4095 - BUFFER_EXTRA;
-        out->allocated = true;
+        oj_out_init_allocate(out);
     }
     out->cur      = out->buf;
     out->circ_cnt = 0;
@@ -600,28 +597,26 @@ void oj_dump_obj_to_json_using_params(VALUE obj, Options copts, Out out, int arg
 }
 
 void oj_write_obj_to_file(VALUE obj, const char *path, Options copts) {
-    char        buf[4096];
-    struct _out out;
-    size_t      size;
-    FILE       *f;
-    int         ok;
+    stack_buffer buf;
+    struct _out  out;
+    size_t       size;
+    FILE        *f;
+    int          ok;
 
-    out.buf       = buf;
-    out.end       = buf + sizeof(buf) - BUFFER_EXTRA;
-    out.allocated = false;
+    oj_out_init_stack_buffer(&out, &buf);
+
     out.omit_nil  = copts->dump_opts.omit_nil;
     oj_dump_obj_to_json(obj, copts, &out);
     size = out.cur - out.buf;
     if (0 == (f = fopen(path, "w"))) {
-        if (out.allocated) {
-            xfree(out.buf);
-        }
+        oj_out_free(&out);
         rb_raise(rb_eIOError, "%s", strerror(errno));
     }
     ok = (size == fwrite(out.buf, 1, size, f));
-    if (out.allocated) {
-        xfree(out.buf);
-    }
+
+    oj_out_free(&out);
+
+    fclose(f);
     if (!ok) {
         int err = ferror(f);
         fclose(f);
@@ -649,18 +644,17 @@ static void write_ready(int fd) {
 #endif
 
 void oj_write_obj_to_stream(VALUE obj, VALUE stream, Options copts) {
-    char        buf[4096];
-    struct _out out;
-    ssize_t     size;
-    VALUE       clas = rb_obj_class(stream);
+    stack_buffer buf;
+    struct _out  out;
+    ssize_t      size;
+    VALUE        clas = rb_obj_class(stream);
 #if !IS_WINDOWS
     int   fd;
     VALUE s;
 #endif
 
-    out.buf       = buf;
-    out.end       = buf + sizeof(buf) - BUFFER_EXTRA;
-    out.allocated = false;
+    oj_out_init_stack_buffer(&out, &buf);
+
     out.omit_nil  = copts->dump_opts.omit_nil;
     oj_dump_obj_to_json(obj, copts, &out);
     size = out.cur - out.buf;
@@ -690,14 +684,11 @@ void oj_write_obj_to_stream(VALUE obj, VALUE stream, Options copts) {
     } else if (rb_respond_to(stream, oj_write_id)) {
         rb_funcall(stream, oj_write_id, 1, rb_str_new(out.buf, size));
     } else {
-        if (out.allocated) {
-            xfree(out.buf);
-        }
+        oj_out_free(&out);
         rb_raise(rb_eArgError, "to_stream() expected an IO Object.");
     }
-    if (out.allocated) {
-        xfree(out.buf);
-    }
+
+    oj_out_free(&out);
 }
 
 void oj_dump_str(VALUE obj, int depth, Out out, bool as_ok) {
@@ -944,6 +935,31 @@ void oj_dump_raw(const char *str, size_t cnt, Out out) {
     assure_size(out, cnt + 10);
     APPEND_CHARS(out->cur, str, cnt);
     *out->cur = '\0';
+}
+
+void oj_out_init_stack_buffer(Out out, stack_buffer * buffer) {
+    out->buf = buffer->buffer;
+    out->end = buffer->buffer + sizeof(buffer->buffer) - BUFFER_EXTRA;
+    out->allocated = false;
+}
+
+void oj_out_init_allocate(Out out) {
+    oj_out_init_allocate_n(out, DEFAULT_BUFFER_SIZE);
+}
+
+void oj_out_init_allocate_n(Out out, size_t cnt) {
+    if (cnt < BUFFER_EXTRA * 2) {
+        cnt += BUFFER_EXTRA;
+    }
+    out->buf = ALLOC_N(char, cnt);
+    out->end = out->buf + cnt - BUFFER_EXTRA;
+    out->allocated = true;
+}
+
+void oj_out_free(Out out) {
+    if (out->allocated) {
+        xfree(out->buf);
+    }
 }
 
 void oj_grow_out(Out out, size_t len) {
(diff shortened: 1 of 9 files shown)


ohler55:
I'll look at the changes more carefully this weekend and also run some benchmarks to verify the extra layer of functions has not impacted performance.

ohler55:
Thank you for suggesting a change. I've got some comments and would like to discuss some points.

First I suppose is the purpose of the PR. I'm not sure what the concern with N +10 is. 10 is just a pad to cover a miscount. It should probably be a `#define` but other than that it is not relied on for anything other than miscounts. Ideally with perfect code it should be 0. In all cases the expected size is calculated first and the buffer is grown before writing to it. If you have encountered a write past the end of the buffer then that's a bug and should be fixed by making sure the calculated size is correct and not bu just bumping up the safety net of 10 bytes to 64.

Ok, the first thing I did was run the benchmarks in `test/perf_object.rb -d -s -m 1000 -i 100` on the existing code and the code in your branch. Your's averaged about 1% or more lower than the existing. I did multiples runs and averaged. I'd prefer not to degrade performance if at all possible unless a bug is fixed.

Looking though the code I noticed that a lot of the change was about replacing one or two lines of code with a function call. For example the `oj_out_free()` function replaces a check for allocated and a free (really 3 if the closing } is counted). I don't think that helps one way or the other in regards to readability and the extra function call may be part of the performance degradation depending on what the compiler does. Another is `oj_out_init_allocate()` which has exactly one line in it. Why bounce the reader around the code when a direct call is more clear and does not add the overhead of calling a function?

Another change I don't feel comfortable with is the stack_buffer type. It seems like a very round about way of creating a fixed size allocation and only serves to obfuscate what the actual intent is. I could see a type that is a buf4k as char[4096] but to wrap it in a struct seems like overkill.

I know that is a lot of commenting and it is not inline but I'd prefer to address the high level issues first before getting into the details. Lets start with the first. What is the reason for the PR? Bug or something else?

rudihorn:
This PR primarily fixes a bug. The oj grow function reads past the boundaries of some stack allocated buffers, and this causes TruffleRuby to crash. The problem with the current code is that you need to verify the correct initialization of the same structure at 8 different call sites, and ensure that each of them correctly matches the specification required by the oj_grow (and its copy in dump_leaf) and other functions. Currently 5 out of those 8 call sites do the wrong thing, and most of the call sites use a hardcoded constant for the extra buffer size.

I'm trying to perform what would be classified as encapsulation in the OOP world. By bundling the allocation, grow and free functions into a single section in the code, it is trivial to understand and verify that any correct sequence of allocate, grow(s) and free will do the right thing. It then becomes easier to verify the correct usage of these three functions at each call site.

If there are performance degradations then one could experiment with inlining such as with the `assure_size` function (https://github.com/ohler55/oj/blob/6f17dda7f482a02cfd6cc175564d79457cb08375/ext/oj/dump.h#L55-L59). This code is semantically doing the same thing, so it should not be changing performance significantly. The PR I submitted may do some things different with initial heap allocation, so it may worth checking if any performance degradation comes from there. There are also some situations where the smaller extra buffer value of 10 may lead to less allocations. It would be good to figure out what the actual extra buffer length should be (but this way there is only a single place to change it). The use of abstraction shouldn't prevent you from retaining any of the performance.

> Another change I don't feel comfortable with is the stack_buffer type. It seems like a very round about way of creating a fixed size allocation and only serves to obfuscate what the actual intent is. I could see a type that is a buf4k as char[4096] but to wrap it in a struct seems like overkill.

The way I see it the reader at the call site doesn't care about the details of the stack buffer. Instead the application should just do the right thing, use the default sized stack buffer and leave the rest of the details to the underlying implementation. You could use a char pointer as well as a length, but this requires you to specify the length of the buffer twice. Defining a struct allows this allocation to be done in a type-safe way (we can use sizeof() in `oj_out_init_stack_buffer`, the call site doesn't have to specify a size). If you just typedef a 4k char buffer, then sizeof() will return the size of a pointer, not the buffer. The struct has no runtime overhead.

ohler55:
You make some reasonable point on encapsulation although the existing code treated `struct _out` as the object. If you don't mind let's leave the encapsulation for a little later and address the memory read violation first. Help me understand what you've seen.

Ok, so you encountered a memory read violation. I don't see how it could be in the oj_grow_out function though as there are no reads in that function other than the memcpy. Is that what you are referring to? If it is can you explain how it is broken and what your changes do to fix it. It's unclear to me how the safety factor of 10 bytes really effects that and if it does where the real issue is.

rudihorn:
Yes it is the `memcpy` instruction which is problematic. A concrete example looks as follows. The programmer allocates a stack buffer using `char buffer[4096]`. In practice our memory values may then be:

```
char buffer[4096] =                                                          0xA000; (arbitrary address, valid until 0xAFFF)
buf->start = buffer =                                                       0xA000;
buf->end = buffer + sizeof(buffer) - 10 = 0xB000 - 10 = 0xAFF6;
// call grow out
memcpy(buf, out->buf, out->end - out->buf + BUFFER_EXTRA);
         //  0xA0000,         0xAFF6 - 0xA000 + 64 = 0x1036
         // reads to 0xB036, which is past 0xAFFF
```

In cases where `buf->end` is allocated as `buffer + sizeof(buffer) - 64`, it works as exected.

ohler55:
I see, so really the first "fix" would be to change fast.c, mimic_json.c, oj.c, rails.c, and string_writer.c to use BUFFER_EXTRA instead of 10.

Second would be to use more shared code. The first fix would likely not cause a change in performance or if it did a look at the proper value for the BUFFER_EXTRA might be considered. When it's clear there is no degradation in performance then step 2 would be making the code cleaner. Maybe with inlines or maybe a macro. You've chosen to create a new object type, if I can call it that for C called stack_buffer. It kind of bleeds over into the Out type. Would it be cleaner if the object for encapsulation was Out instead of stack_buffer?

Anyway, to proceed would you be okay with a 2 step fix? I know it's a bit more work for you but it would make it easier to identify where the performance drop is.

rudihorn:
For me the mentioned snippet was producing more noise than any accurate measurements of degradation is able to produce. I've split the PR up into two commits, the first only replaces any `10` constants with `BUFFER_EXTRA`, and the second adds the rest of the changes. Maybe you can see if you can get a more accurate measurement of any performance changes. So you should be able to checkout each commit and then set BUFFER_EXTRA accordingly.

> Would it be cleaner if the object for encapsulation was Out instead of stack_buffer?
The stack_buffer should be an internal type of the out type. I'm not entirely sure what you are suggesting concretely though.

ohler55:
I'll try to put together some better benchmarks this weekend.

ohler55:
If you made 2 PRs I'd be glad to merge the first one while we discuss the second. I agree the code should be more clean though so as long as you can tolerate the delays lets explore some variations.

ohler55:
The change to use the BUFFER_EXTRA of 64 instead of 10 didn't seem to make a significant difference in performance. Using the other changes did add around 1% overhead. Not a lot but I think it could be zero. I'll describe in the next comment.

Here is the script I used for benchmarking each branch:

```
#!/usr/bin/env ruby
# encoding: UTF-8

$: << '.'
$: << File.join(File.dirname(__FILE__), "../lib")
$: << File.join(File.dirname(__FILE__), "../ext")

require 'optparse'
require 'oj'

$verbose = false
$indent = 2
$iter = 100_000
$size = 2

opts = OptionParser.new
opts.on("-v", "verbose")                                    { $verbose = true }
opts.on("-c", "--count [Int]", Integer, "iterations")       { |i| $iter = i }
opts.on("-i", "--indent [Int]", Integer, "indentation")     { |i| $indent = i }
opts.on("-s", "--size [Int]", Integer, "size (~Kbytes)")    { |i| $size = i }
opts.on("-h", "--help", "Show this display")                { puts opts; Process.exit!(0) }
files = opts.parse(ARGV)

$obj = {
  'a' => 'Alpha', # string
  'b' => true,    # boolean
  'c' => 12345,   # number
  'd' => [ true, [false, [-123456789, nil], 3.9676, ['Something else.', false], nil]], # mix it up array
  'e' => { 'zero' => nil, 'one' => 1, 'two' => 2, 'three' => [3], 'four' => [0, 1, 2, 3, 4] }, # hash
  'f' => nil,     # nil
  'h' => { 'a' => { 'b' => { 'c' => { 'd' => {'e' => { 'f' => { 'g' => nil }}}}}}}, # deep hash, not that deep
  'i' => [[[[[[[nil]]]]]]]  # deep array, again, not that deep
}

Oj.default_options = { :indent => $indent, :mode => :strict }

if 0 < $size
  o = $obj
  $obj = []
  (4 * $size).times do
    $obj << o
  end
end

$json = Oj.dump($obj)
GC.start
start = Time.now
$iter.times { Oj.dump($obj) }
duration = Time.now - start
puts "Dumped #{$json.length} byte JSON #{$iter} times in %0.3f seconds or %0.3f iteration/sec." % [duration, $iter / duration]
```

ohler55:
First let me say that I know this might be asking too much from you and if so let me know and I can make changes in the direction you suggested. I do appreciate what you have done and are doing so please don't take it wrong. Anyway here what I had in mind..

The `struct _out` has the pointers to the buffer being used and it is really the "object" in a OOP approach. How about adding the 4096 array to that structure. It keeps all the parts together without creating a new stack_buffer. If you feel strongly about the naming of the grow function and would prefer to use "assure" do that directly without another call in between. Maybe an `oj_out_init()` would take care of the setup.

One thing I have learned over the decade since Oj was originally written is that conditionals have more overhead that a function call in some cases. I don't think it will make a difference overall but a free function pointer could be used instead of the allocated flag. Not really part of this PR but I might give it a try after.

ohler55:
Thanks for splitting the MR. I'll get the first one merged right away.

Yes, adding buffer to the struct out is what I was referring to. `oj_str_writer_init()` should be able to use the out buffer until it grows too large. I'm not sure why it was a special case.

Not sure about the whether the tweak for branch prediction would make a difference or not. In any case the impact would be very small.

ohler55:
Do you plan on taking my suggestion for using the struct out instead of stack_buffer?

ohler55:
I'll look this evening. It looks the same as before though unless I missed something.

rudihorn:
The code now doesn't have a `stack_buffer` type anymore and only has a single initialization function `oj_out_init`. But otherwise hasn't changed compared to the stack_refactor branch of mine.

ohler55:
Looks good. Github was a bit strange showing the changes but checking out the branch and looking at the code was fine. Merging.

ohler55:
Thanks for your patience.

rudihorn:
No problem, thanks for merging this!

PR #758, opened 2022-05-08
Fix use-after-free in file pointer

Watson1978:
GCC 12 reports use-after-free warning at compilation.

```
dump.c: In function ‘oj_write_obj_to_file’:
dump.c:627:19: warning: pointer ‘f’ may be used after ‘fclose’ [-Wuse-after-free]
  627 |         int err = ferror(f);
      |                   ^~~~~~~~~
dump.c:625:5: note: call to ‘fclose’ here
  625 |     fclose(f);
      |     ^~~~~~~~~
```

This patch will fix this warning.


ohler55:
Thanks. Good catch. The right tooling helps. :-)

Issue #770, opened 2022-06-08
max_depth option behaviour changed in Oj >= v3.13.10


ggeverae:
:wave: Hi! Thanks for sharing this gem!

We're running Oj on Ruby `ruby 2.7.6` together with Rails `6.1.6`. We run `Oj.optimize_rails` in an initializer and proceed to parse JSON data with `ActiveSupport::JSON.decode` further down in the app.

When we try upgrading from v3.13.9, `JSON::NestingError: Too deeply nested`  gets thrown in tests/production where there wasn't any problem before.

We've finally found the time to perform some detailed investigations and this is what came up:

### Change of behaviour in Oj >= v3.13.10

In #720 the option parsing code has been refactored in `oj/ext/oj/mimic_json.c` from:

```c
mimic_parse_core() {
  // ...
  pi.max_depth             = 100;

  if (Qnil != ropts) {
    //...

    // fetch oj_max_nesting_sym option from options hash
    v = rb_hash_lookup(ropts, oj_max_nesting_sym);
    if (Qtrue == v) {
      pi.max_depth = 100;
    } else if (Qfalse == v || Qnil == v) {
      // set max_depth to 0 if option was false, 'nil' or option was omitted
      pi.max_depth = 0;
    } else if (T_FIXNUM == rb_type(v)) {
      pi.max_depth = NUM2INT(v);
    }
    //...
  }
  //...
}
```

to

```c
parse_options_cb() {
  //...

  // check if options hash specifies oj_max_nesting_sym option
  } else if (oj_max_nesting_sym == k) {
    // if it does, override default
    if (Qtrue == v) {
        pi->max_depth = 100;
    } else if (Qfalse == v || Qnil == v) {
        pi->max_depth = 0;
    } else if (T_FIXNUM == rb_type(v)) {
        pi->max_depth = NUM2INT(v);
    }
  }
  //...
}

mimic_parse_core() {
  // ...
  pi.max_depth             = 100;

  if (Qnil != ropts) {
    //...
    rb_hash_foreach(ropts, parse_options_cb, (VALUE)&pi);
    //...
  }
  //...
}
```

The difference in behaviour thus being that calling `JSON.parse` with some options hash not containing a `max_nesting` key would originally reset `max_depth` to 0, whereas now it's kept at the default 100.

[...]

PR #773, opened 2022-06-11
Revert max_nesting behavior when max_nesting option was not given

Watson1978:
Fix https://github.com/ohler55/oj/issues/770

Revert max_nesting behavior to v3.13.9, however, this behavior is not compatible with Ruby standard JSON gem.


ohler55:
I have to agree that it seems wrong but rails is full of cases that seem wrong to many of us. Thanks for responding and handling this.

Issue #772, opened 2022-06-10
Oj.load_file can't open files which names have accented chars


cvjxendra:
Windows 10, 64 bits

>  Oj.load_file 'Conceição1.json'

[1] pry> Oj.load_file 'Conceição1.json'
=> nil
[2] pry> File.exists? 'Conceição1.json'
=> true

ohler55:
I'll look into it.

ohler55:
I don't have a Windows machine to test on but I can say that it worked on MacOS which indicates there is some Windows specific unicode string handling. Are you able to open the file and then pass the file to Oj.load? Does that work as expected?

cvjxendra:
>  Are you able to open the file and then pass the file to Oj.load? 

If I rename 'Conceição1.json' to 'Conceicao1.json', Oj works.

PR #774, opened 2022-06-12
Convert file path to wide char to handle unicode file path on Windows

Watson1978:
Seems that it need to convert file path to wide char to handle unicode file path on Windows using [MultiByteToWideChar ](https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar)

rb_w32_mbstr_to_wstr() will convert it well.

Fix https://github.com/ohler55/oj/issues/772


ohler55:
You are truly awesome! Thanks for the support.

PR #776, opened 2022-06-19
Warn on implicit Oj::Rails.mimic_JSON

skryukov:
Implicit mode change might affect `Oj.dump` / `Oj.load`, so I propose to warn user on this change:

```ruby
require 'oj'

Oj.generate 'hello'
# => warning: Oj::Rails.mimic_JSON was called implicitly. Call it explicitly beforehand if you want to remove this warning.
```


ohler55:
Sounds reasonable.

PR #777, opened 2022-06-20
Implement `:escape_mode => :slash`

casperisfine:
Similar to https://github.com/flori/json/pull/405

It's a "cheap" way to make JSON safe to interpolate in a `<script>` tag.


ohler55:
Looks okay to me.

lucianosousa:
@ohler55 do you mind in create a new version with this change?
I would love to get it in my project without forcing the git SHA :)

Thanks!

PR #790, opened 2022-07-15
Use SSE instruction to copy the data of structure

Watson1978:
When compiled with gcc, the data is copied by the `rep movsq` instruction at the point where the structure is assigned and the 8 bytes are copied at a time.

```
       │    parse_info_init(&pi);
       │    pi.options   = oj_default_options;
  1.29 │      mov  oj_default_options@@Base-0x1748,%rsi
  0.12 │      mov  $0x2b,%ecx
       │      lea  0x10f8(%rsp),%rdi
 82.95 │      rep  movsq %ds:(%rsi),%es:(%rdi)
```

(The perf command shows the above result in my environment.)

The SSE instruction can be used to increase the size of the copy at a time.

−       | before | after  | result
--       | --     | --     | --
Oj.load  | 4.975M | 5.215M | 1.048x

### Environment
- Linux
  - Manjaro Linux x86_64
  - Kernel: 5.18.0-1-rt11-MANJARO
  - AMD Ryzen 7 5700G
  - gcc version 12.1.0
  - Ruby 3.1.2

### Before
```
             Oj.load   495.911k i/100ms
             Oj.load      4.975M (± 0.2%) i/s -     50.087M in  10.068529s
```

### After
```
             Oj.load   521.707k i/100ms
             Oj.load      5.215M (± 0.2%) i/s -     52.171M in  10.004304s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

json = "42"

Benchmark.ips do |x|
  x.time = 10

  x.report('Oj.load') do |times|
    i = 0
    while i < times
      Oj.load(json)
      i += 1
    end
  end
end
```

### DIFF
--- a/ext/oj/compat.c
+++ b/ext/oj/compat.c
@@ -10,6 +10,7 @@
 #include "parse.h"
 #include "resolve.h"
 #include "trace.h"
+#include "util.h"
 
 static void hash_set_cstr(ParseInfo pi, Val kval, const char *str, size_t len, const char *orig) {
     const char *   key    = kval->key;
@@ -225,7 +226,7 @@ oj_compat_parse(int argc, VALUE *argv, VALUE self) {
     struct _parseInfo pi;
 
     parse_info_init(&pi);
-    pi.options              = oj_default_options;
+    oj_memcpy(&pi.options, &oj_default_options, sizeof(struct _options));
     pi.handler              = Qnil;
     pi.err_class            = Qnil;
     pi.max_depth            = 0;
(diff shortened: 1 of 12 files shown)


Watson1978:
Seems that original memcpy() is faster than my implementation...

```
             Oj.load   548.465k i/100ms
             Oj.load      5.476M (± 0.2%) i/s -     54.846M in  10.015495s
```

Watson1978:
clang outputs highly optimized binaries, so it will slow down on clang env if applying this patch.

I'll investigate this for a while.

Watson1978:
By the way, the article on https://danlark.org/2020/10/08/i-need-extra-c-c-performance-now-how/ is interesting

ohler55:
Very informative article, thanks.

Issue #789, opened 2022-07-14
oj gem 3.13.15+ causes an illegal instruction on CPUs that do not support SSE4.2


Supermathie:
After the introduction of `cmpestri` in a455c389c0bd09cb591d1144e9366fc3f3280da0, people with CPUs that do not support SSE4.2 cannot use the gem.

We have multiple reports of ruby abending with an Illegal Instruction from users of Discourse:

https://meta.discourse.org/t/discourse-update-keeps-failing/231862/25?u=supermathie

https://meta.discourse.org/t/discourse-update-keeps-failing/231862/33?u=supermathie

Can we add a runtime check here to ensure the instruction is supported prior to using it?

ohler55:
@Watson1978 can you look into this?

Watson1978:
it might be solved by https://github.com/ohler55/oj/pull/791.
if not, it might to need to remove the codes for SSE.

ohler55:
@Supermathie can you give the develop branch a try?

Supermathie:
I can't, since I don't actually have hardware lacking SSE4.2 support.

But for *us* (and I suspect for others who do the same) this won't work in the long term without additional workarounds and continual awareness of this since we ship a docker image with compiled gems.

A workaround would be having the user compile & install the gem on startup to ensure appropriate SSE4.2 [non]usage. Not ideal, but doable.

Watson1978:
If we're going to check at runtime, I think it's better to delete the relevant codes because of the overhead...

ohler55:
How about an environment variable that is checked in extconf.rb to block use of SSE4.2?

SamSaffron:
> If we're going to check at runtime, I think it's better to delete the relevant codes because of the overhead...

Can't we compile to variants and then at first load of the library fiddle with function pointers so it points at the right place, thus erasing all the runtime overhead and only having this be minor boot time overhead?

eduardobonet:
We are also seeing many reports of this upon upgrading to 3.13.16: https://gitlab.com/gitlab-org/gitlab/-/issues/368656

ohler55:
@Watson1978 maybe the SSE4.2 code needs to be remove (commented out) until these issue are resolved or maybe just make it a manually activate option if a specific environment variable is set during compile.

eduardobonet:
Can it be reverted as mitigation?

stanhu:
[...]

PR #791, opened 2022-07-15
Enable the SSE4.2 flag only when the compiled executable file works properly

Watson1978:
https://github.com/ohler55/oj/issues/789

SSE4.2 instructions will not be used when it causes SEGV during configuration.


ohler55:
I like it. Very creative solution.

ohler55 (approved):

PR #794, opened 2022-07-25
Provide a way to disable SSE 4.2 instructions

stanhu:
We often ship gems compiled on a SSE 4.2 machine, but some users run
on older hardware. To avoid an illegal instruction error, disable SSE
4.2 instructions by setting the `eOJ_NO_SSE4_2` environment variable.

Relates to https://github.com/ohler55/oj/issues/789

### DIFF
--- a/ext/oj/extconf.rb
+++ b/ext/oj/extconf.rb
@@ -47,7 +47,7 @@ int main(int argc, char **argv) {
 }
 SRC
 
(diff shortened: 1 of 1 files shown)


stanhu:
I'm inclined to suggest that the original merge request be reverted and converted to a runtime check.

stanhu:
#795 might be better.

PR #795, opened 2022-07-25
Perform runtime check for SSE 4.2 instructions

stanhu:
If SSE 4.2 is not available on the system, don't attempt to use SIMD instructions.

Relates to https://github.com/ohler55/oj/issues/789

### DIFF
--- a/ext/oj/parse.c
+++ b/ext/oj/parse.c
@@ -192,7 +192,26 @@ static inline const char *scan_string_noSIMD(const char *str, const char *end) {
     return str;
 }
 
-#if defined(OJ_USE_SSE4_2)
+// Taken from Tensorflow:
+// https://github.com/tensorflow/tensorflow/blob/5dcfc51118817f27fad5246812d83e5dccdc5f72/tensorflow/core/lib/hash/crc32c_accelerate.cc#L21-L38
+#ifdef __SSE4_2__
+#if defined(__x86_64__) && defined(__GNUC__) && \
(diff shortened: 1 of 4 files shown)


ext/oj/parse.c
```diff
@@ -213,6 +213,25 @@ static inline const char *scan_string_SIMD(const char *str, const char *end) {
 }
 #endif
 
+static bool cpu_supports_sse42() {
+#if __has_builtin(__builtin_cpu_supports)
```
stanhu, line 217: Something's not working here. Hang on.

ohler55:
Detecting earlier is better. Running the check on each parse is expensive. I'd rather the test was done when the Makefile is generated. There is a check in the current extconf.rb. Is that not working? If not I could use some help getting that to work on some of the older machines you might have access to. Is that possible?

stanhu:
> Detecting earlier is better. Running the check on each parse is expensive. 

Doesn't this change only do the detection when the module is initialized?

> I'd rather the test was done when the Makefile is generated. There is a check in the current extconf.rb. Is that not working?

We package a build using SSE 4.2 platform, but some users run the code on older hardware. We either need a way to disable the SSE instructions outright, or make the check dynamic.

ohler55:
I see, looks like what you have should be good then. I'll merge once tests are done and cut a release.

ohler55 (approved):

ohler55:
Released

PR #802, opened 2022-07-29
Remove unnessary volatile qualifier in Oj.load

Watson1978:
When we use volatile qualifier, seems it has some performance penalty on GCC.
The volatile qualifier is unnecessary since the VALUE type object has been kept on the local stack.
(If we manage VALUE on the heap, we need to be careful.)

−       | before  | after   | result
--       | --      | --      | --
GCC      | 4.969M  | 5.165M  | 1.039x
Clang    | 5.054M  | 5.090M  | -

### Environment
- Linux
  - Manjaro Linux x86_64
  - Kernel: 5.18.12-3-MANJARO
  - AMD Ryzen 7 5700G
  - gcc version 12.1.0
  - clang version 14.0.6
  - Ruby 3.1.2

#### Before
```
             Oj.load   497.783k i/100ms
             Oj.load      4.969M (± 0.1%) i/s -     49.778M in  10.017042s
```

#### After
```
             Oj.load   518.127k i/100ms
             Oj.load      5.165M (± 0.1%) i/s -     51.813M in  10.031075s
```

#### Before
```
             Oj.load   503.712k i/100ms
             Oj.load      5.054M (± 0.1%) i/s -     50.875M in  10.067027s
```

#### After
```
             Oj.load   510.215k i/100ms
             Oj.load      5.090M (± 0.1%) i/s -     51.022M in  10.022973s
```

### Test code
```ruby
require 'benchmark/ips'
require 'oj'

json = Oj.dump(42)

Benchmark.ips do |x|
  x.warmup = 5
  x.time = 10

  x.report('Oj.load') do |times|
    i = 0
    while i < times
      Oj.load(json)
      i += 1
    end
  end
end
```

### DIFF
--- a/ext/oj/parse.c
+++ b/ext/oj/parse.c
@@ -667,7 +667,7 @@ static void read_num(ParseInfo pi) {
 }
 
 static void array_start(ParseInfo pi) {
-    volatile VALUE v = pi->start_array(pi);
+    VALUE v = pi->start_array(pi);
 
     stack_push(&pi->stack, v, NEXT_ARRAY_NEW);
 }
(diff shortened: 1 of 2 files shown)


ohler55:
I'd have to see more tests with large files that cause a GC while parsing. My understanding is the volatile keyword cause the variable not to be put in a register since Ruby does not"see" register variables. Making the VALUE volatile tests the compiler it may change to put it on the stack instead of in a register.

Watson1978:
Added the test

ohler55:
I'll have to remember `GC.stress`. Looks very useful when testing.

ohler55 (approved):

Issue #796, opened 2022-07-27
`require 'oj'` on TruffleRuby 22.2.0 causes `Polyglot::ForeignException`


okuramasafumi:
```
irb(main):3.0.3:003:0> require 'oj'
/Users/okuramasafumi/.gem/truffleruby/3.0.3/gems/oj-3.13.18/ext/oj/parse.c:246:in `oj_scanner_init': External LLVMFunction __cpu_indicator_init cannot be found. (Polyglot::ForeignException)
	from /Users/okuramasafumi/.gem/truffleruby/3.0.3/gems/oj-3.13.18/ext/oj/oj.c:1753:in `Init_oj'
	from <internal:/Users/okuramasafumi/.rubies/truffleruby-22.2.0/lib/mri/rubygems/core_ext/kernel_require.rb>:64:in `<unknown>'
	from <internal:/Users/okuramasafumi/.rubies/truffleruby-22.2.0/lib/mri/rubygems/core_ext/kernel_require.rb>:64:in `Kernel#gem_original_require'
	from <internal:/Users/okuramasafumi/.rubies/truffleruby-22.2.0/lib/mri/rubygems/core_ext/kernel_require.rb>:64:in `Kernel#require'
	from /Users/okuramasafumi/.gem/truffleruby/3.0.3/gems/oj-3.13.18/lib/oj.rb:13:in `<top (required)>'
	from <internal:core> core/kernel.rb:234:in `Kernel#gem_original_require'
	from <internal:/Users/okuramasafumi/.rubies/truffleruby-22.2.0/lib/mri/rubygems/core_ext/kernel_require.rb>:139:in `Kernel#require'
	from (irb):1:in `<top (required)>'
	from <internal:core> core/kernel.rb:407:in `Kernel#loop'
	from <internal:core> core/throw_catch.rb:36:in `Kernel#catch'
	from <internal:core> core/throw_catch.rb:36:in `Kernel#catch'
	from /Users/okuramasafumi/.rubies/truffleruby-22.2.0/lib/gems/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
	from <internal:core> core/kernel.rb:376:in `Truffle::KernelOperations.load'
	from <internal:core> core/kernel.rb:376:in `Kernel#load'
	from /Users/okuramasafumi/.rubies/truffleruby-22.2.0/bin/irb:42:in `<main>'
```

On TruffleRuby version 22.2.0, `require 'oj'` causes `Polyglot::ForeignException`.
[...]

PR #803, opened 2022-07-30
Add TruffleRuby CI job

stanhu:
This should help catch future Truffleruby regressions, such as #796.

Note that `truffleruby-head` hit a number of seg faults, while `truffleruby+graalvm` seems to work fine.


stanhu:
Closing in favor of #804.

PR #804, opened 2022-07-30
Add TruffleRuby into CI

Watson1978:
This PR will add TruffleRuby CI lane.
The failing tests are simply skipped now.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -78,3 +78,22 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
       - run: bundle exec rake
+
+  TruffleRuby:
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [ '22.2.0' ]
+
+    env:
+      BUNDLE_GEMFILE: gemfiles/no_rails.gemfile
+      MAKE: make -j 2
+    runs-on: ubuntu-latest
+    name: TruffleRuby ${{ matrix.ruby }} (ubuntu)
+    steps:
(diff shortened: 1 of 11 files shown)


ohler55 (approved):

stanhu:
@Watson1978 FYI, I had better luck with `truffleruby+graalvm` over `truffleruby-head` in #803. I'm not sure why, but perhaps @eregon has some ideas how to fix the seg fault and other issues that manifested in the tests.

.github/workflows/CI.yml
```diff
@@ -78,3 +78,22 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
       - run: bundle exec rake
+
+  TruffleRuby:
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [ '22.2.0' ]
```
eregon, line 86: I think it would be better to test the latest truffleruby release, so just `truffleruby` here and `ruby-version: ${{ matrix.ruby }}` below (or just no matrix).

eregon (commented):
Thank you for adding TruffleRuby in CI!
Looks like there are quite a few failures, will track them in https://github.com/oracle/truffleruby/issues/2701

eregon:
> @Watson1978 FYI, I had better luck with `truffleruby+graalvm` over `truffleruby-head` in #803.

That's a release vs latest dev build. I expect it should be about the same between  `truffleruby+graalvm` over `truffleruby` (both releases, both run in native mode, the default).

eregon:
Most skips removed in https://github.com/ohler55/oj/pull/817

PR #806, opened 2022-08-02
Disable SSE 4.2 compile instructions by default

stanhu:
Compiling with `-msse4.2` is problematic because it may cause a number
of other SSE instructions to be used. For example:

```
$ ./elfx86exts oj.so
MODE64 (call)
CMOV (cmovne)
SSE2 (movq)
SSE41 (pinsrq)
SSE1 (movups)
SSSE3 (pshufb)
SSE3 (movddup)
SSE42 (pcmpestri)
CPU Generation: Unknown
```

The runtime check in #795 is therefore not sufficient to prevent SSE
instructions from leaking elsewhere in the shared library. As a
result, loading Oj on older machines with these libraries will result
in a "illegal instruction" crash.

Introduce a `--with-sse42` flag that can be used to configure this
gem.

### DIFF
--- a/ext/oj/extconf.rb
+++ b/ext/oj/extconf.rb
@@ -34,21 +34,9 @@ have_func('rb_hash_bulk_insert', 'ruby.h') unless '2' == version[0] && '6' == ve
 
 dflags['OJ_DEBUG'] = true unless ENV['OJ_DEBUG'].nil?
 
-src =<<~SRC
-#include <string.h>
-#include <nmmintrin.h>
-int main(int argc, char **argv) {
-    const char *str = "hello               ";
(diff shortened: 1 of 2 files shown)


stanhu:
@Watson1978 Please take a look. I tried to make runtime detection work, but it is a bit challenging to keep the shared library from only using SSE instructions for the specific optimization.

Watson1978:
Unfortunately...
However I think this is better.

ohler55:
Looks like you both agree this is the best approach. If so I'll merge and release.

stanhu:
@ohler55 Great, thank you.

ohler55 (approved):

PR #810, opened 2022-08-17
Time fix

aardvark179:
While investigating the time related failures on TruffleRuby I noticed that Unix times with offsets do not actually round trip correctly on MRI either. This PR creates times by using `Time.at` with nanoseconds and an offset, and avoids the precision issues of the old method. I've also added an assertion to the tests that the loaded object is equal to the one originally dumped out.


ohler55:
Looks like an invalid call was used.

aardvark179:
Ah, can't use `rb_funcallv_kw` on 2.5. Let's see how we do with a time creation function that definitely is in 2.5.

ohler55 (approved):

PR #811, opened 2022-08-22
Range fix

aardvark179:
Remove the assumption that `Range` objects are `T_STRUCTS`.


ohler55:
Please use the clang formatter on the changes. That should un-tabify the changes you made so the formatting is consistent. Otherwise it looks fine.

ohler55:
I see the Truffle CI is failing. Thanks for working on this by the way.

aardvark179:
Ah. One of the test needs a fix for TruffleRuby itself which hasn't been merged yet. I've restored that skip for now, and will remove it once the TruffleRuby PR has merged.

ohler55 (approved):

PR #812, opened 2022-08-22
Add some null checks to mark functions.

aardvark179:
Added some extra null checks for issues found when running very mark functions very aggressively while running OJ tests.

### DIFF
--- a/ext/oj/fast.c
+++ b/ext/oj/fast.c
@@ -677,21 +677,23 @@ static void free_doc_cb(void *x) {
 }
 
 static void mark_leaf(Leaf leaf) {
-    switch (leaf->value_type) {
-    case COL_VAL:
-        if (NULL != leaf->elements) {
-            Leaf first = leaf->elements->next;
-            Leaf e     = first;
+    if (NULL != leaf) {
+        switch (leaf->value_type) {
+        case COL_VAL:
+            if (NULL != leaf->elements) {
+                Leaf first = leaf->elements->next;
+                Leaf e     = first;
 
-            do {
-                mark_leaf(e);
-                e = e->next;
-            } while (e != first);
+                do {
+                    mark_leaf(e);
+                    e = e->next;
+                } while (e != first);
+            }
+            break;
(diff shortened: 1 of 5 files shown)


ext/oj/fast.c
```diff
@@ -677,21 +677,23 @@ static void free_doc_cb(void *x) {
 }
 
 static void mark_leaf(Leaf leaf) {
-    switch (leaf->value_type) {
-    case COL_VAL:
-        if (NULL != leaf->elements) {
-            Leaf first = leaf->elements->next;
-            Leaf e     = first;
+    if (NULL != leaf) {
```
ohler55, line 680: Was there a code change here or just reformatting? Please use the clang formatter in any case.

aardvark179: There is a code change, that initial check that `leaf` isn't null. Clang format leaves my change unaffected, but does touch quite a lot of other lines in that file.

ohler55:
How does the use of DATA_PTR impact performance? Any change?

The NULL check parts are good though.

ext/oj/fast.c
ohler55: Odd, the .clang-format file indicates to never use tabs yet I think the diff shows the introduction of tabs unless I am mistaken.

aardvark179: I think it's just showing the different in white space (adding surrounding `if (NULL != leaf` increases the indentation level).

aardvark179:
The added `DATA_PTR` should have a minimal impact on performance. On MRI it's one indirection, and on TruffleRuby it's the only clue we have that you might have accessed the wrapped struct, and that it might need to have its mark function run again.

Ultimately I'd like to submit a patch that avoids using mark functions entirely (because they are relatively expensive to emulate on TruffleRuby), but that still needs a little work.

ohler55:
Interesting to learn that Truffle Ruby tracks by use and not by marking.

I see what you mean about the indentation. I suspect you are correct. I'll approve once the tests complete.

ohler55 (approved):

PR #818, opened 2022-10-22
Move GC verification to end of process in tests

Watson1978:
Since Ruby 3.2.0, `GC.verify_compaction_references` has the change.
https://github.com/ruby/ruby/commit/a6dd859affc42b667279e513bb94fb75cfb133c1

If `expand_heap` is present, It will certainly try to expand the heap area on each calling.
We are calling it many time on `teardown`, and then Ruby process try to allocate a large amount of memory.
When a large amount of memory space is allocated, it takes a very long time to expand it at next time.
Finally, the test fails with a timeout on CI.

I think it would be sufficient to perform GC validation on survived object at end of process.
This change should eliminate failures on Ruby 3.2 CI lane.

### DIFF
--- a/test/test_object.rb
+++ b/test/test_object.rb
@@ -221,13 +221,6 @@ class ObjectJuice < Minitest::Test
 
   def teardown
     Oj.default_options = @default_options
-#=begin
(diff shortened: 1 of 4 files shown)


ohler55:
That make sense.

ohler55 (approved):

PR #819, opened 2022-10-27
Inject start and end byte offsets of objects into Hash

meinac:
Hi @ohler55 👋 

This change is probably far away from being acceptable but I wanted to open the PR instead of opening an issue to discuss things on a concrete example.

At GitLab we are dealing with big JSON artifacts(10GB is a regular size) regularly and this is becoming a bottleneck. So I had the idea of storing the byte offsets of some important JSON objects in the database(PG, Redis, etc.) to download and parse only the relevant parts of the JSON instead of downloading the whole artifact and parsing it from scratch. 

The vendors we are using supports downloading parts of the JSON artifacts with [HTTP range header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) so if we can somehow get the byte offsets of each object, we will be able to do this.

I've already prepared this patch which embeds the `start_byte` and `end_byte` of the JSON object into generated Ruby Hash as instance variables. How do you feel about this approach?

Thanks in advance for any support you can offer 🙏

### DIFF
--- a/ext/oj/parser.c
+++ b/ext/oj/parser.c
@@ -1163,6 +1163,7 @@ static void parser_mark(void *ptr) {
 extern void oj_set_parser_validator(ojParser p);
 extern void oj_set_parser_saj(ojParser p);
 extern void oj_set_parser_usual(ojParser p);
+extern void oj_set_parser_introspected(ojParser p);
 extern void oj_set_parser_debug(ojParser p);
 
 static int opt_cb(VALUE rkey, VALUE value, VALUE ptr) {
@@ -1240,8 +1241,10 @@ static VALUE parser_new(int argc, VALUE *argv, VALUE self) {
                 oj_set_parser_validator(p);
             } else if (0 == strcmp("debug", ms)) {
                 oj_set_parser_debug(p);
+            } else if (0 == strcmp("introspected", ms)) {
+                oj_set_parser_introspected(p);
             } else {
-                rb_raise(rb_eArgError, "mode must be :validate, :usual, :saj, or :object");
+                rb_raise(rb_eArgError, "mode must be :validate, :usual, :saj, :object, :debug, or :introspected");
             }
         }
         if (1 < argc) {
@@ -1482,6 +1485,29 @@ static VALUE parser_usual(VALUE self) {
     return usual_parser;
 }
 
+static VALUE introspected_parser = Qundef;
+
+/* Document-method: introspected
+ * call-seq: introspected
+ *
+ * Returns the default introspected parser. Note the default introspected parser can not be
+ * used concurrently in more than one thread.
+ */
+static VALUE parser_introspected(VALUE self) {
+    if (Qundef == introspected_parser) {
+        ojParser p = ALLOC(struct _ojParser);
+
+        memset(p, 0, sizeof(struct _ojParser));
+        buf_init(&p->key);
+        buf_init(&p->buf);
+        p->map = value_map;
+        oj_set_parser_introspected(p);
+        introspected_parser = Data_Wrap_Struct(parser_class, parser_mark, parser_free, p);
+        rb_gc_register_address(&introspected_parser);
+    }
+    return introspected_parser;
+}
+
 static VALUE saj_parser = Qundef;
 
 /* Document-method: saj
@@ -1552,6 +1578,7 @@ void oj_parser_init(void) {
     rb_define_method(parser_class, "method_missing", parser_missing, -1);
 
     rb_define_module_function(parser_class, "usual", parser_usual, 0);
+    rb_define_module_function(parser_class, "introspected", parser_introspected, 0);
     rb_define_module_function(parser_class, "saj", parser_saj, 0);
     rb_define_module_function(parser_class, "validate", parser_validate, 0);
 }
--- a/ext/oj/usual.c
+++ b/ext/oj/usual.c
@@ -43,7 +43,7 @@ typedef union _key {
     };
     struct {
         int16_t xlen;  // should be the same as len
-        char *  key;
+        char   *key;
     };
 } * Key;
 
@@ -54,9 +54,9 @@ typedef union _key {
 #define BYTE_OFFSETS_STACK_INC_SIZE 256
 
 // Holds the start byte offsets of each JSON object encountered
-typedef struct byte_offsets_S{
-    int length;
-    int current;
+typedef struct _byte_offsets {
+    int   length;
+    int   current;
     long *stack;
 } ByteOffsets;
 
@@ -83,7 +83,7 @@ typedef struct _delegate {
     VALUE array_class;
     VALUE hash_class;
 
-    char *  create_id;
+    char   *create_id;
     uint8_t create_id_len;
     uint8_t cache_str;
     uint8_t cache_xrate;
@@ -154,8 +154,8 @@ static VALUE resolve_classname(VALUE mod, const char *classname, bool auto_defin
 static VALUE resolve_classpath(const char *name, size_t len, bool auto_define) {
     char        class_name[1024];
     VALUE       clas;
-    char *      end = class_name + sizeof(class_name) - 1;
-    char *      s;
+    char       *end = class_name + sizeof(class_name) - 1;
+    char       *s;
     const char *n = name;
 
     clas = rb_cObject;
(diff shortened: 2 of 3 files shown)


ohler55:
Have you considered using the callback parsers? SAJ or better the new parser series. It will be easier to make a custom parser with the start and end offsets with the new new parser.

meinac:
Thanks for your quick response @ohler55 🙏 The SAJ parser and other callback parsers will probably require too much changes on our end as we need the whole JSON to be parsed but the byte offsets embedded.

Do you mean something like implementing a new mode like `wab` by new parser series? If so, I can give it a try.

ohler55:
Take a look at parser.c as well as usual.c and saj2.c. In particular the usual.c. You can probably derive what you are looking for from that.

meinac:
Thanks for giving me the direction @ohler55! I've added a new parser called `introspected` which injects a key called `__oj_introspection` to each Ruby hash generated along with the start and end byte offsets.

The performance of this new parser is not that good as the usual parser as expected but still better than the stdlib;

```
             OJ load   166.950k i/100ms
        introspected   133.279k i/100ms
               usual   242.627k i/100ms
              stdlib    96.920k i/100ms
             OJ load      1.669M (± 1.4%) i/s -      8.348M in   5.003789s
        introspected      1.423M (± 4.6%) i/s -      7.197M in   5.066559s
               usual      2.437M (± 3.9%) i/s -     12.374M in   5.084624s
              stdlib    982.066k (± 1.1%) i/s -      4.943M in   5.033786s

Comparison:
               usual:  2437145.2 i/s
             OJ load:  1668564.2 i/s - 1.46x  (± 0.00) slower
        introspected:  1423331.2 i/s - 1.71x  (± 0.00) slower
              stdlib:   982065.6 i/s - 2.48x  (± 0.00) slower
```

I've added my changes to `usual.c` as otherwise I would have to either copy all the content of it, or create another utility file to include in both parsers.

Would you mind checking this version of the PR?

ohler55:
Looks like a good start although there are a few things I'd like to see changed.

For consistency please following the conventions I use in the rest of the code for struct names. Also run the clang-format.

I would also like to keep usual.c to be just the usual parser. That will require a few changes so that it can be shared with the introspect parser. If the usual delegate is renamed to Usual or UsualDelegate and put in a header usual.h then introspect.c can embed the struct _usual in in struct _introspect.

If you want me to hop in and make some changes to either your PR or in some other branch I'd be glad to do something like that.

I'm a little disappointed that the performance isn't slower to the usual parser. I wonder if there is something that could be done to make it faster. Maybe a different approach for collecting offsets. I didn't notice how the offsets were being used. If you can describe a bit more about how the offsets would be used once collected that might help me make a suggestion.

Anyway, you are on the right track in my opinion.

ohler55:
Odd that clang-format with the .clang-format file changed the original code. I wonder if something changed in the latest version or something.

Anyway, thanks for that. Can we change "introspected" to 'introspect" so all modes are current tense and not past? Then the more challenging part of moving to a new file. Would you like me to do that or are you okay with that?

meinac:
Thanks a lot for your feedback @ohler55 🙏 

> I would also like to keep usual.c to be just the usual parser. That will require a few changes so that it can be shared with the introspect parser. If the usual delegate is renamed to Usual or UsualDelegate and put in a header usual.h then introspect.c can embed the struct _usual in in struct _introspect.

The new introspected parser would share almost 99% of the logic implemented in `usual.c` so would you like me to encapsulate the common logic in a utility file which can be then included in both `usual.c` and `introspected.c`?

> If you want me to hop in and make some changes to either your PR or in some other branch I'd be glad to do something like that.

Since I'm not writing C in my daily routine that would be a huge help if you can actually make changes on this branch 🙏 

> I'm a little disappointed that the performance isn't slower to the usual parser. I wonder if there is something that could be done to make it faster. Maybe a different approach for collecting offsets.

Actually collecting the offsets is not adding too much overhead but attaching those values to the Hash objects is where the difference is coming from. I've commented out the `set_introspection_values` function to test this and it works at the same speed as the `usual` parser.

> I didn't notice how the offsets were being used. If you can describe a bit more about how the offsets would be used once collected that might help me make a suggestion.

Here is a bit more background information about why I need the byte offsets;

At GitLab's threat insights team, we are using the JSON artifacts generated by CI jobs that contain a project's vulnerability list to inform the users about the vulnerabilities they have in their projects. These JSON artifacts can have a size of 10GB and can contain tens of thousands of vulnerabilities. We had the plan to ingest those vulnerabilities into the PostgreSQL database to support filtering, sorting, and pagination without loading the whole document into memory but some properties of these vulnerabilities can be huge and given we can have billions(yes billions) of records, storing them in the database can waste expensive disk space of our database. So instead of storing those properties in the database, if I store their byte offsets, then whenever needed I can read only the required range of bytes from the JSON artifact to access the data and parse it as those will be valid JSON fragments.

Here I have a PoC MR on GitLab: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/101517/diffs

- [Here](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/101517/diffs#ebf4f24f12f868c4d799ddc49515ea75d1a89b53_76_89) where I save the byte offsets of `vulnerability#remediation` objects along with the vulnerability records
- [Here](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/101517/diffs#dc9b017ebaae19ac8bcd239a75574d462d0ae39b_98_143) where I read the JSON artifact partially to construct the `vulnerability#remediation` without downloading and parsing the whole JSON file.

---

Actually, introspected feels better as an option for all parsers but I am not sure if there will be too many people using it :)

ohler55:
I had a thought but don't know if it will work. If I make changes to Oj so that the usual can be seen by third party (you) code. I wonder if introspect can be an extension to Oj.

Anyway, do you want to add me as a contributor to your branch so I can make direct changes or should I make a PR for your PR. :-)

ohler55:
Alternatively I could make a branch off of Oj itself.

meinac:
> I had a thought but don't know if it will work. If I make changes to Oj so that the usual can be seen by third party (you) code. I wonder if introspect can be an extension to Oj.

I thought the same and actually, that would be a really nice thing so that other people can also write extensions for OJ.

> do you want to add me as a contributor to your branch so I can make direct changes or should I make a PR for your PR. :-)

Thanks a lot @ohler55 🙏 I've invited you to my clone so you can push changes directly to this branch.

ohler55:
I created a branch called parser-extension that has changes that I think will allow third party extensions.

First call oj_parser_new(). The return value will be a VALUE that wraps the parser. Next create your introspect(ed) delegate. The delegate struct should have struct _usual as a member. Call oj_init_usual(ojParser p, Usual d). This will set up the usual delegate. Next replace as needed:

```c
    p->ctx    = (void *)d; // point to the introspect delegate
    p->option = option;
    p->result = result;
    p->free   = dfree; // replace this
    p->mark   = mark; // replace this
    p->start  = start;
```

The other parser fields may not need to be modified. The p->funcs[TOP_FUN].open_object should be replaced but keep the previous value to call before or after setting the location. Same for the close_object function.

If you can get a separate project to link and build that would be a great start. You will need to add a ruby function to do the steps described above.

meinac:
It's amazing! I was able to implement a new parser(within the project for testing purposes) easily!

I think did a bit too much indirection to wrap the original functions but can't think of any better way. Would you mind checking my implementation if the direction is correct? https://github.com/meinac/oj/compare/parser-extension...implement-introspect-extension

ohler55:
Right idea but try creating a separate repo or project. You should be able to do that without modifying Oj. If you create the project I'll help get it working.

meinac:
> Or I could create a separate project if you prefer.

Both options work for me but first I have to figure out how to link the parser extension to the OJ gem. If you already have a concrete plan, then I would say it's better if you create the project so I can just write C functions.

ohler55:
I'll create a project tomorrow. If we can't figure out how to link it in we can fall back to modifying Oj.

meinac:
Hi @ohler55 👋 I was able to compile and run a seperate project by setting the compiler flags, here is how I set my `extconf.rb` file;

```ruby
require "mkmf"
require "bundler"

oj_gemspec = Bundler.definition.specs.find { _1.name == "oj" }

LIBDIR      = RbConfig::CONFIG['libdir']
INCLUDEDIR  = RbConfig::CONFIG['includedir']
OJ_HEADERS  = oj_gemspec.full_gem_path + "/ext/oj"

HEADER_DIRS = [INCLUDEDIR, OJ_HEADERS]

LIB_DIRS = [LIBDIR]

dir_config('stree', HEADER_DIRS, LIB_DIRS)

create_makefile("oj-introspect_parser/introspect_parser")
```

ohler55:
Nice. I tried just setting up a reference to an external function and that looked like it might work. Looks like you are doing even better. If you can add me to the repo I can contribute later this evening.

ohler55:
We can try it on your project once we get it working and passing a few tests.

I'll make some changes to the setup later to allow for options to be processed along with startup and free to work correctly.

PR #823, opened 2022-11-04
Fix new parser interface

meinac:
Looks like changing the struct members after calling `Data_Wrap_Struct` is causing segfaults. We can solve this by changing the method signature of `oj_parser_new` to inject the dependency(parser special init function).

The current bug can be reproduced by the following steps on the `oj-introspect` project;

- Pull the latest changes
- Run `./bin/console`
- Execute the following on console: `50_000.times { Oj::Introspect.new.parse('{}') }`

### DIFF
--- a/ext/oj/parser.c
+++ b/ext/oj/parser.c
@@ -1259,7 +1259,7 @@ static VALUE parser_new(int argc, VALUE *argv, VALUE self) {
 // from this function. A delegate must be added before the parser can be
 // used. Optionally oj_parser_set_options can be called if the options are not
 // set directly.
-VALUE oj_parser_new() {
+VALUE oj_parser_new(void(*init_function)(ojParser parser)) {
     ojParser p = ALLOC(struct _ojParser);
 
 #if HAVE_RB_EXT_RACTOR_SAFE
@@ -1270,6 +1270,8 @@ VALUE oj_parser_new() {
     buf_init(&p->key);
     buf_init(&p->buf);
     p->map = value_map;
+
+    init_function(p);
 
     return Data_Wrap_Struct(parser_class, parser_mark, parser_free, p);
 }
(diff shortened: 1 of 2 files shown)


ohler55:
Nice catch on the failure but I'd like to take a different approach to solving it. The issue is that the `mark()` function is being called in one thread and the parser.ctx is changed in another thread. The fix should be to make parser.ctx volatile and also put a mutex around changes in ctx and the section of the mark that reference the parser.ctx.

I'll get a fix in this weekend unless you want to give it a try. It would involve adding a function called `parser_set_ctx` and the additions of a mutex like this bit of code:
```c
#ifdef HAVE_PTHREAD_MUTEX_INIT
    pthread_mutex_unlock(&stack->mutex);
#else
    rb_mutex_unlock(stack->mutex);
#endif
```

meinac:
> The issue is that the mark() function is being called in one thread and the parser.ctx is changed in another thread.

@ohler55 - This was running on a single thread. Do you mean that, apart from this issue, we can also have segfaults for multi-thread runs even if we merge this MR?

> I'll get a fix in this weekend unless you want to give it a try.

🙏

ohler55:
The GC is a separate thread I believe.

ohler55:
I did a little more exploring. I don't think we need the mutex. The GVL should take care of that. What I did discover is that GC is not being triggered when the allocated bytes reaches the limit. Instead Ruby crashes. If the test is `50_000.times { Oj::Introspect.new.parse('{}'); GC.start }` there is no crash.

Now to figure out how to work around that without continuing to call GC.stat.

ohler55:
BTW, creating a new parser each time instead of reusing the old one is not ideal. The intent is to reuse the parser. Having said that I still need to figure out why Ruby is not triggering GC when allocated exceeds the limit.

ohler55:
Ah ha. An alloc in usual.c is called which triggers a GC which calls the parser mark function.

ohler55:
Try the "mark-it" branch.

meinac:
It works 🎉

ohler55:
v3.13.23 includes the fix.

PR #836, opened 2023-01-22
Fix SEGV if to_s method returns invalid object

Watson1978:
Fix https://github.com/ohler55/oj/issues/834

In Ruby, it can easily override the `to_s` method and it may not return a string as expected (rarely).
This PR will check the object using `StringValue`  whether it is a string object as expected after calling `to_s`.

### DIFF
--- a/ext/oj/oj.h
+++ b/ext/oj/oj.h
@@ -379,6 +379,12 @@ extern bool oj_use_hash_alt;
 extern bool oj_use_array_alt;
 extern bool string_writer_optimized;
 
+static inline VALUE oj_safe_string_convert(VALUE obj) {
+    VALUE rstr = rb_funcall(obj, oj_to_s_id, 0);
+    StringValue(rstr);
+    return rstr;
+}
+
 #define APPEND_CHARS(buffer, chars, size) \
     { \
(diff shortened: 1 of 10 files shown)


ext/oj/oj.h
```diff
@@ -379,6 +379,12 @@ extern bool oj_use_hash_alt;
 extern bool oj_use_array_alt;
 extern bool string_writer_optimized;
 
+static inline VALUE oj_safe_string_convert(VALUE obj) {
+    VALUE rstr = rb_funcall(obj, oj_to_s_id, 0);
```
ohler55, line 383: Should also verify the to_s method exists.

Watson1978:
```ruby
obj = BasicObject.new
p Oj.dump(obj, mode: :compat)
#=> json.rb:4:in `dump': undefined method `to_s' for #<BasicObject:0x0000559fe073d748> (NoMethodError)
```

`rb_funcall` raises a `NoMethodError` exception if `to_s` method does not exist.
I think this is expected behavior.

ohler55: True, it is good enough.

ohler55 (approved):

PR #851, opened 2023-02-27
Format Markdown files with Markdownlint

nschonni:
Took a quick pass over the files with Markdownlint-CLI2. Still 2 left over in docs/index.md for HTML elements.

Basic config of
```json
{
    "config": {
        "default": true,
        "MD013": false
    }
}
```

Leaves the last 2 messages:

```
$ markdownlint-cli2-fix "**/*.md"
markdownlint-cli2-fix v0.6.0 (markdownlint v0.27.0)
Finding: **/*.md
Linting: 22 file(s)
Summary: 2 error(s)
docs/index.md:3:1 MD033/no-inline-html Inline HTML [Element: a]
docs/index.md:3:41 MD033/no-inline-html Inline HTML [Element: img]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@
 ## 3.13.8 - 2021-09-27
 
 - Fix `Oj::Doc` behaviour for inexisting path.
+
   ```ruby
   Oj::Doc.open('{"foo":1}') do |doc|
     doc.fetch('/foo/bar') # used to give `1`, now gives `nil`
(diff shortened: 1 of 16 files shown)


ohler55:
The docs directory looks like an experiment that didn't pan out so it can be removed.

nschonni:
I think those files are for the GitHub pages https://ohler55.github.io/oj/
Do you want me to remove the files and the `_config.yml` in the root? You'd probably need to turn off the Pages in the repo settings too

ohler55:
I see, should probably leave them in place until I have a changes to see if GitHub Pages is being used or not.

PR #855, opened 2023-02-28
Update minimum requirement to Ruby 2.7

nschonni:
Splitting this out of the other Rubocop PR just to see if it can land with the clean Truffleruby run without the changes


ohler55 (approved):

.github/workflows/CI.yml
```diff
@@ -21,31 +21,22 @@ jobs:
...
-          - rails_6
           - rails_6.1
           - rails_7
         exclude:
           - os: macos
             ruby: head
-          - ruby: head
+        include:
+          - ruby: '2.7'
             gemfile: rails_5
```
nschonni, line 37: Looks like 5 is actually EOL as well https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html
Should it get dropped too?

ohler55: sure

PR #861, opened 2023-03-02
Enable Windows Rails CI

nschonni:
I'm expecting this to fail as I was testing on an earlier branch before.
I figured out one of the failure methods from going through the first error message that pointed to https://github.com/tzinfo/tzinfo/wiki/Resolving-TZInfo::DataSourceNotFound-Errors

Opening to see if the other failure gives anyone else an idea

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,10 +34,6 @@ jobs:
             ruby: head
           - os: windows
             ruby: head
-          - os: windows
-            gemfile: rails_6.1
-          - os: windows
-            gemfile: rails_7
         include:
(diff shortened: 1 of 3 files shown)


nschonni:
You can see in the failure that it's not picking up the correct timezone still
```
 Failure:
TestJSONEncoding#test_time_to_json_includes_local_offset [D:/a/oj/oj/test/activesupport6/encoding_test.rb:129]:
--- expected
+++ actual
@@ -1 +1 @@
-"\"2005-02-01T15:15:10.000-05:00\""
+"\"2005-02-01T15:15:10.000+00:00\""
```

PR #862, opened 2023-03-02
Autofix Rubocop Minitest - Batch 4

nschonni:
Autofix the no-option Minitest rules


nschonni:
The random TruffleRuby only failure might actually be catching something. The check_circular does appear to have a parameter that using `assert_predicate` is a little more explicit in checking

ohler55:
No longer relevant.

PR #866, opened 2023-03-13
Cleanup RUBY_INTEGER_UNIFICATION

Watson1978:
RUBY_INTEGER_UNIFICATION was introduced at Ruby 2.4.0. https://github.com/ruby/ruby/blob/8c6b349805e2f17a57576b8dfad31e5681d6b0e9/doc/NEWS/NEWS-2.4.0#compatibility-issues-excluding-feature-bug-fixes-

Now, oj gem supports Ruby 2.7.0 or later.
https://github.com/ohler55/oj/blob/d289f4243b0bebd0ceb01c19cc8c338aff0c54d9/oj.gemspec#L21

So, this patch will remove unnecessary code for old Ruby.


ohler55 (approved):

ohler55:
It seems like only yesterday that integer unification happened. Thanks you

PR #868, opened 2023-03-18
Remove redundant rb_check_type for String

Watson1978:
StringValuePtr() is called after checking the type, but StringValuePtr() also performs type checking.

StringValuePtr() will invoke rb_string_value() which performs Implicit type conversion and type checking.

```
VALUE
rb_string_value(volatile VALUE *ptr)
{
    VALUE s = *ptr;
    if (!RB_TYPE_P(s, T_STRING)) {
	s = rb_str_to_str(s);
	*ptr = s;
    }
    return s;
}
```


ohler55 (approved):

ohler55:
Sorry, I forgot about this one.

Watson1978:
Thanks 😊

PR #869, opened 2023-03-19
Fix JSON.create_id= to allow nil

Watson1978:
`JSON.create_id=` method accepts `nil` as an argument in document. https://github.com/ohler55/oj/blob/2c1a622660ed5ab94b9fdfbcc7bc16017307972a/ext/oj/mimic_json.c#L655

However, it raises TypeError when passed `nil`.

```
irb(main):001:0> require 'oj'
=> true
irb(main):002:0> Oj.mimic_JSON
=> JSON
irb(main):003:0> JSON.create_id = nil
(irb):3:in `create_id=': wrong argument type nil (expected String) (TypeError)
        from (irb):3:in `<main>'                             
        from /home/watson/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
        from /home/watson/.rbenv/versions/3.2.1/bin/irb:25:in `load'
        from /home/watson/.rbenv/versions/3.2.1/bin/irb:25:in `<main>'
irb(main):004:0> 
```

This patch will accept `nil` as an argument.

### DIFF
--- a/ext/oj/mimic_json.c
+++ b/ext/oj/mimic_json.c
@@ -654,11 +654,9 @@ static VALUE mimic_recurse_proc(VALUE self, VALUE obj) {
  *
  * - *id* [_nil_|String] new create_id
  *
- * Returns [_String_] the id.
+ * Returns [_nil_|_String_] the id.
  */
 static VALUE mimic_set_create_id(VALUE self, VALUE id) {
-    Check_Type(id, T_STRING);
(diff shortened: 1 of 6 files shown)


ohler55 (approved):

ohler55:
arigato

PR #873, opened 2023-04-06
Remove dead code in oj_hash_init

peterzhu2118:


ohler55:
I believe classes can move hence the need for the registration.

peterzhu2118:
`rb_gc_register_address` takes a pointer to a `VALUE`, but `cache_class` is allocated on the stack which means that it becomes invalid memory the moment the function returns. You can see the [implementation here](https://github.com/ruby/ruby/blob/bffadcd6d46ccfccade79ce0efb60ced8eac4483/gc.c#L9209).

This function doesn't seem to be using `cache_class` in any way that is unsafe for compaction, so we don't need to pin it (i.e. tell compaction that this object cannot move).

ohler55:
You are correct. Thanks.

ohler55 (approved):

PR #876, opened 2023-04-22
Remove checking whether rb_gc_mark_movable exists

Watson1978:
Currently the oj gem supports Ruby 2.7+.

rb_gc_mark_movable() has introduced at Ruby 2.7.
Ref. https://github.com/ruby/ruby/blob/v2_7_0/include/ruby/intern.h#L540

So, we can use rb_gc_mark_movable() freely in oj gem.

### DIFF
--- a/ext/oj/extconf.rb
+++ b/ext/oj/extconf.rb
@@ -24,7 +24,6 @@ dflags = {
 }
 
 # Support for compaction.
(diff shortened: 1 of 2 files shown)


ohler55 (approved):

ohler55:
39

PR #881, opened 2023-05-28
Move development dependency gems into Gemfile to fix lint error

Watson1978:
This patch will fix lint error in https://github.com/ohler55/oj/actions/runs/5104692030/jobs/9175781272


Watson1978:
Hmmm, looks like I need a little research.

PR #885, opened 2023-06-03
Fix some Rubocop error

Watson1978:


ohler55:
Let's take a different approach. Please no todo file. Fixing the files is better.

ohler55:
The rest of the suggested changes can be made.

Looking good so far. I will be away from a computer for 2 days. I will look when I get back.

PR #879, opened 2023-05-07
Optimize integer value dumps

Watson1978:
To convert an Integer to a String, it needs to do division and remainder calculations many times. These calculations can consume more CPU cycles than other instructions.

This patch will prepare a lookup table which has the remainder (0~99) of a number divided by 100 and use the table to omit the calculation.

−       | before  | after   | result
--       | --      | --      | --
Oj.dump  | 3.816M  | 4.255M  | 1.115x

### Environment
- Linux
  - Manjaro Linux x86_64
  - Kernel: 6.3.0-1-MANJARO
  - AMD Ryzen 7 5800H
  - gcc version 12.2.1
  - Ruby 3.2.2

### Before
```
Warming up --------------------------------------
             Oj.dump   386.110k i/100ms
Calculating -------------------------------------
             Oj.dump      3.816M (± 1.4%) i/s -     19.306M in   5.060327s
```

### After
```
Warming up --------------------------------------
             Oj.dump   425.525k i/100ms
Calculating -------------------------------------
             Oj.dump      4.255M (± 0.2%) i/s -     21.276M in   5.000424s
```

### Test code
```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'oj'
end

data = [646086033, 414841692, 706653378, 1069473884, 181209966, 515120040, 892957102, 689595306, 719771732, 651396679, 549722480]

Benchmark.ips do |x|
  x.report('Oj.dump') { Oj.dump(data) }
end
```

PR #886, opened 2023-06-10
Refactor converting number to string by using optimized function

Watson1978:
Refactor converting number to string by using optimized function which introduced at https://github.com/ohler55/oj/pull/879

### DIFF
--- a/ext/oj/code.c
+++ b/ext/oj/code.c
@@ -185,24 +185,17 @@ void oj_code_attrs(VALUE obj, Attr attrs, int depth, Out out, bool with_class) {
             } else {
                 char   buf[32];
                 char  *b   = buf + sizeof(buf) - 1;
-                int    neg = 0;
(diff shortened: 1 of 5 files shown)


ohler55 (approved):

ohler55:
Before merging, is the performance the same or better with this change?

Watson1978:
Sorry for late reply.

When I run benchmark with same code in https://github.com/ohler55/oj/pull/879,
it is 0.8% slower by this PR.

I believe it is a tolerance error.

## Before
```
             Oj.dump   431.211k i/100ms
             Oj.dump      4.323M (± 0.7%) i/s -     21.992M in   5.087412s
```

### After
```
             Oj.dump   425.450k i/100ms
             Oj.dump      4.288M (± 0.3%) i/s -     21.698M in   5.060483s
```

PR #887, opened 2023-07-02
Refactor in where obtain hash keys by using common function

Watson1978:
Refactor in where obtain hash keys by using common function. However, it might have the margin of error for the overhead of calling oj_calc_hash_key().

−       | before   | after    | result
--       | --       | --       | --
Oj.dump  | 449.521k | 447.738k | 0.996x

### Environment
- Linux
  - Manjaro Linux x86_64
  - Kernel: 6.3.5-2-MANJARO
  - AMD Ryzen 7 5800H
  - gcc version 13.1.1
  - Ruby 3.2.2

### Before
```
             Oj.load    44.738k i/100ms
             Oj.load    449.521k (± 0.9%) i/s -      6.755M in  15.029388s
```

### After
```
             Oj.load    44.117k i/100ms
             Oj.load    447.738k (± 0.3%) i/s -      6.750M in  15.075672s
```

### Test code
```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'oj'
end

json =<<-EOF
{
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
    },
    "lastName": {
      "type": "string",
    },
    "age": {
      "type": "integer",
      "minimum": 0
    }
  }
}
EOF

Benchmark.ips do |x|
  x.warmup = 5
  x.time = 15

  x.report('Oj.load') { Oj.load(json, mode: :compat) }
end
```

### DIFF
--- a/ext/oj/compat.c
+++ b/ext/oj/compat.c
@@ -13,32 +13,19 @@
 #include "trace.h"
 
 static void hash_set_cstr(ParseInfo pi, Val kval, const char *str, size_t len, const char *orig) {
-    const char    *key    = kval->key;
-    int            klen   = kval->klen;
-    Val            parent = stack_peek(&pi->stack);
-    volatile VALUE rkey   = kval->key_val;
+    const char *key    = kval->key;
+    int         klen   = kval->klen;
+    Val         parent = stack_peek(&pi->stack);
 
-    if (Qundef == rkey && Yes == pi->options.create_ok && NULL != pi->options.create_id &&
+    if (Qundef == kval->key_val && Yes == pi->options.create_ok && NULL != pi->options.create_id &&
         *pi->options.create_id == *key && (int)pi->options.create_id_len == klen &&
         0 == strncmp(pi->options.create_id, key, klen)) {
         parent->classname = oj_strndup(str, len);
         parent->clen      = len;
     } else {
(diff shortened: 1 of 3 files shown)


ext/oj/compat.c
```diff
@@ -25,20 +25,8 @@ static void hash_set_cstr(ParseInfo pi, Val kval, const char *str, size_t len, c
         parent->clen      = len;
     } else {
         volatile VALUE rstr = oj_cstr_to_value(str, len, (size_t)pi->options.cache_str);
+        rkey                = oj_calc_hash_key(pi, kval);
```
ohler55, line 28: A further optimization would be to remove line 19 then line 21 becomes:
`if (Qundef == val->key_val && Yes == pi->options.create_ok && NULL != pi->options.create_id &&`
and the new line 28 becomes `volatile VALUE rkey                = oj_calc_hash_key(pi, kval);`.

Since oj_calc_hash_key already gets rkey from the parent the suggested change avoids the assignment and stack variable us in the first block.

Watson1978: Thanks for your review. I have fixed the codes to follow the suggestion.

ohler55 (approved):

PR #909, opened 2024-01-07
Clean up HAVE_RB_HASH_BULK_INSERT

Watson1978:
`rb_hash_bulk_insert` function can be used as public API since Ruby 2.7.0. https://github.com/ruby/ruby/commit/d3df725f9efeb0e6c121f9a1316abaa90956f1b8

Current Oj gem only supports Ruby 2.7.0 or later.
So there is no need to check whether function exists.


ohler55 (approved):

ohler55:
Thanks you.

PR #910, opened 2024-01-09
Clean up HAVE_RB_GC_MARK_MOVABLE

Watson1978:
rb_gc_mark_movable has introduced at Ruby 2.7.0
https://github.com/ruby/ruby/commit/aac4d9d6c7e6b6b0742f3941b574f6006ccb5672

Current Oj gem only supports Ruby 2.7.0 or later.
So there is no need to check whether function exists.

### DIFF
--- a/ext/oj/extconf.rb
+++ b/ext/oj/extconf.rb
@@ -26,7 +26,6 @@ dflags = {
 }
 
 # Support for compaction.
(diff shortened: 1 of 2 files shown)


Watson1978:
I forgot that truffleruby doesn't have rb_gc_mark_movable :(

Issue #911, opened 2024-01-31
Rejecting invalid UTF-8 without opting in to XSS character escaping


KJTsanaktsidis:
Oj is blindly serialising strings with invalid unicode sequences in them that I expected it would reject, when using Rails mode and setting `ActiveSupport::JSON::Encoding.escape_html_entities_in_json = false`. I'm happy for literal `<`, `&`, etc characters to appear in my JSON, but I do _not_ want to emit JSON with invalid UTF-8 bytes!

Oj correctly mimics the JSON gem's default behaviour of throwing an exception here:

```
irb(main):001> require 'json'
=> true
irb(main):002> broken_string = "very\xAEbroken"
=> "very\xAEbroken"
irb(main):003> JSON.dump broken_string
/Users/ktsanaktsidis/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/json-2.7.1/lib/json/common.rb:305:in `generate': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
irb(main):004> require 'oj'
=> true
irb(main):005> Oj.mimic_JSON
=> JSON
irb(main):006> JSON.dump broken_string
(irb):6:in `dump': Invalid Unicode [ae 62 72 6f 6b] at 4 (JSON::GeneratorError)
```

So this is good. However, it's not correctly mimicking ActiveSupports' behaviour:

```
irb(main):001> require 'active_support/all'
=> true
irb(main):002> broken_string = "very\xAEbroken"
=> "very\xAEbroken"
irb(main):003> ActiveSupport::JSON::Encoding.escape_html_entities_in_json = true
=> true
irb(main):004> broken_string.to_json
/Users/ktsanaktsidis/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/json-2.7.1/lib/json/common.rb:305:in `generate': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
irb(main):005> ActiveSupport::JSON::Encoding.escape_html_entities_in_json = false
=> false
irb(main):006> broken_string.to_json
/Users/ktsanaktsidis/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/json-2.7.1/lib/json/common.rb:305:in `generate': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
irb(main):007> require 'oj'
=> true
irb(main):010> Oj.optimize_rails
=> nil
[...]

PR #912, opened 2024-02-01
Make invalid Unicode data raise when encoding through Oj::Rails::Encoder

KJTsanaktsidis:
This is a potential fix for https://github.com/ohler55/oj/issues/911. Currently, whether or not `Oj::Rails::Encoder` raises on invalid unicode data depends on the value of `ActiveSupport.escape_html_entities_in_json`. In order to accurately mimic the behaviour of stock Rails with the stock json gem, it should in fact raise an exception regardless.

I've so far deliberately copied rather than shared functionality that's shared between `RailsEsc` and `RailsXEsc` mode, because I wasn't quite sure how to factor the similarities out. We can leave it like this, or I'm happy to take pointers on a way to factor this down better.

I added a testcase for invalid Unicode to the Rails 6 & 7 encoding tests, and also parameterised the existing unicode-related tests to make sure they work correctly with both settings of `ActiveSupport.escape_html_entities_in_json`.

### DIFF
--- a/ext/oj/dump.c
+++ b/ext/oj/dump.c
@@ -198,7 +198,18 @@ inline static long rails_xss_friendly_size(const uint8_t *str, size_t len) {
 }
 
 inline static size_t rails_friendly_size(const uint8_t *str, size_t len) {
-    return calculate_string_size(str, len, rails_friendly_chars);
+    long    size = 0;
+    size_t  i    = len;
+    uint8_t hi   = 0;
+
+    for (; 0 < i; str++, i--) {
+        size += rails_friendly_chars[*str];
+        hi |= *str & 0x80;
+    }
+    if (0 == hi) {
+        return size - len * (size_t)'0';
+    }
+    return -(size - len * (size_t)'0');
 }
 
 const char *oj_nan_str(VALUE obj, int opt, int mode, bool plus, int *lenp) {
@@ -750,8 +761,9 @@ void oj_dump_raw_json(VALUE obj, int depth, Out out) {
 void oj_dump_cstr(const char *str, size_t cnt, bool is_sym, bool escape1, Out out) {
     size_t      size;
     char       *cmap;
-    const char *orig   = str;
-    bool        has_hi = false;
+    const char *orig                  = str;
+    bool        has_hi                = false;
+    bool        do_unicode_validation = false;
 
     switch (out->opts->escape_mode) {
     case NLEsc:
@@ -772,8 +784,9 @@ void oj_dump_cstr(const char *str, size_t cnt, bool is_sym, bool escape1, Out ou
         size = xss_friendly_size((uint8_t *)str, cnt);
         break;
     case JXEsc:
-        cmap = hixss_friendly_chars;
-        size = hixss_friendly_size((uint8_t *)str, cnt);
+        cmap                  = hixss_friendly_chars;
+        size                  = hixss_friendly_size((uint8_t *)str, cnt);
+        do_unicode_validation = true;
         break;
     case RailsXEsc: {
         long sz;
@@ -786,12 +799,22 @@ void oj_dump_cstr(const char *str, size_t cnt, bool is_sym, bool escape1, Out ou
         } else {
             size = (size_t)sz;
         }
+        do_unicode_validation = true;
         break;
     }
-    case RailsEsc:
+    case RailsEsc: {
+        long sz;
         cmap = rails_friendly_chars;
-        size = rails_friendly_size((uint8_t *)str, cnt);
+        sz   = rails_friendly_size((uint8_t *)str, cnt);
+        if (sz < 0) {
+            has_hi = true;
+            size   = (size_t)-sz;
+        } else {
+            size = (size_t)sz;
+        }
+        do_unicode_validation = true;
         break;
+    }
     case JSONEsc:
     default: cmap = hibit_friendly_chars; size = hibit_friendly_size((uint8_t *)str, cnt);
     }
@@ -822,7 +845,7 @@ void oj_dump_cstr(const char *str, size_t cnt, bool is_sym, bool escape1, Out ou
         for (; str < end; str++) {
             switch (cmap[(uint8_t)*str]) {
             case '1':
-                if ((JXEsc == out->opts->escape_mode || RailsXEsc == out->opts->escape_mode) && check_start <= str) {
+                if (do_unicode_validation && check_start <= str) {
                     if (0 != (0x80 & (uint8_t)*str)) {
                         if (0xC0 == (0xC0 & (uint8_t)*str)) {
                             check_start = check_unicode(str, end, orig);
@@ -846,8 +869,7 @@ void oj_dump_cstr(const char *str, size_t cnt, bool is_sym, bool escape1, Out ou
                 }
                 break;
             case '3':  // Unicode
-                if (0xe2 == (uint8_t)*str && (JXEsc == out->opts->escape_mode || RailsXEsc == out->opts->escape_mode) &&
-                    2 <= end - str) {
+                if (0xe2 == (uint8_t)*str && do_unicode_validation && 2 <= end - str) {
                     if (0x80 == (uint8_t)str[1] && (0xa8 == (uint8_t)str[2] || 0xa9 == (uint8_t)str[2])) {
                         str = dump_unicode(str, end, out, orig);
                     } else {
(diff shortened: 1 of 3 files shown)


ohler55:
I'll have to spend a little time looking at the changes. It make me uncomfortable that some of the existing tests were removed. Did you run any benchmarks to see what impact on performance the change had?

KJTsanaktsidis:
I definitely shuffled some tests around but I didn’t mean to remove any. What did I delete? I’ll fix that for sure :)

good call on benchmarking - I’ll put something together today.

ohler55:
Maybe it was the moving around part that made it appear as if some tests were removed. I will look more carefully.

ohler55:
Were you able to put together a benchmark and fix the clang formatting issues?

KJTsanaktsidis:
Sorry, I haven’t gotten to that - I do plan to in the next couple of days!

KJTsanaktsidis:
OK, I've fixed the clang-format problems, and I put together a benchmark: https://gist.github.com/KJTsanaktsidis/f85be084d61aca54f8493ab63fe0707f

Without this patch:
```
long_7bit_printable_string with RailsXEsc mode
                          2.133k (± 4.0%) i/s -     10.835k in   5.088119s
long_7bit_printable_string with RailsEsc mode
                          1.914k (± 3.3%) i/s -      9.750k in   5.099691s
long_ascii_string with RailsXEsc mode
                        163.294 (± 1.8%) i/s -    832.000 in   5.096756s
long_ascii_string with RailsEsc mode
                        175.841 (± 2.3%) i/s -    884.000 in   5.029591s
long_angle_bracket_string with RailsXEsc mode
                        275.981 (± 2.9%) i/s -      1.400k in   5.077551s
long_angle_bracket_string with RailsEsc mode
                          1.951k (± 3.3%) i/s -      9.945k in   5.103185s
long_utf8_multibyte_string with RailsXEsc mode
                        359.297 (± 2.2%) i/s -      1.800k in   5.012476s
long_utf8_multibyte_string with RailsEsc mode
                        654.515 (± 2.3%) i/s -      3.315k in   5.067395s
```

With this patch:
```
long_7bit_printable_string with RailsXEsc mode
                          2.120k (± 3.0%) i/s -     10.761k in   5.080555s
long_7bit_printable_string with RailsEsc mode
                          2.130k (± 3.0%) i/s -     10.812k in   5.081807s
long_ascii_string with RailsXEsc mode
                        169.397 (± 2.4%) i/s -    848.000 in   5.009189s
long_ascii_string with RailsEsc mode
                        179.139 (± 2.2%) i/s -    901.000 in   5.032772s
long_angle_bracket_string with RailsXEsc mode
                        284.548 (± 2.5%) i/s -      1.430k in   5.028596s
long_angle_bracket_string with RailsEsc mode
                          2.119k (± 3.1%) i/s -     10.750k in   5.079096s
long_utf8_multibyte_string with RailsXEsc mode
                        425.998 (± 2.3%) i/s -      2.142k in   5.031145s
long_utf8_multibyte_string with RailsEsc mode
                        426.126 (± 2.3%) i/s -      2.142k in   5.029241s
```

The only substantial difference is that the "lots of multibyte characters" case with this patch now takes the same amount of time regardless of using `RailsEsc` mode or `RailsXEsc` mode, whereas before it was faster in `RailsEsc` mode (because it wasn't validating any of the characters). But in non-multibyte-heavy workloads it seems the same.

ohler55:
Benchmarks look good. I'll start a more detailed review to get this merged.

test/activesupport6/encoding_test.rb
```diff
@@ -74,42 +74,77 @@ def test_hash_keys_encoding
...
-    # set to true, the default, before running the test. This might be wrong but
-    # for now it will have to do.
-    ActiveSupport.escape_html_entities_in_json = true
-    result = ActiveSupport::JSON.encode("€2.99")
-    assert_equal '"€2.99"', result
-    assert_equal(Encoding::UTF_8, result.encoding)
-
-    result = ActiveSupport::JSON.encode("✎☺")
-    assert_equal '"✎☺"', result
-    assert_equal(Encoding::UTF_8, result.encoding)
```
ohler55, line 90: The test functions in this file were taken directly from the ActiveSupport code base. I did add the comment and the one line described. Is there a reason to remove the test?

ohler55: I see that it was not removed but moved and the comment deleted. Is the comment no longer valid?

KJTsanaktsidis: I moved it into the `UnicodeTests` module so that it could be included and run twice - once with `escape_html_entities_in_json` on and once with it off - https://github.com/ohler55/oj/pull/912/files/eb3febb2e03389b573ba1144eb301a0629a7de5a#diff-5919a44f303dc6030d533e858a18ac8298c483aa397f31c89cf465e09ef5447aR134

KJTsanaktsidis: > I see that it was not removed but moved and the comment deleted. Is the comment no longer valid?

Yes - the module the test is included in now explicitly sets `escape_html_entities_in_json` (to both values) in a `setup` block. So it passes with that setting both on and off, so I don't think the comment is necessary anymore.

ext/oj/dump.c
```diff
@@ -786,12 +799,22 @@ void oj_dump_cstr(const char *str, size_t cnt, bool is_sym, bool escape1, Out ou
...
+        long sz;
         cmap = rails_friendly_chars;
-        size = rails_friendly_size((uint8_t *)str, cnt);
+        sz   = rails_friendly_size((uint8_t *)str, cnt);
+        if (sz < 0) {
+            has_hi = true;
+            size   = (size_t)-sz;
+        } else {
+            size = (size_t)sz;
+        }
```
ohler55, line 814: Under what circumstances could `sz` ever be less than 0?

test/activesupport6/encoding_test.rb
ohler55: Right but the point of the comment was to let someone looking at the test know. what the modifications were to the original ActiveSupport tests. Given that you should probably note the module was set up to set that value which is different than the original tests. I'm happy to have the module there but it does need a comment explaining why it is there.

ext/oj/dump.c
ohler55: I suspect your fork might be behind. That's probably the issue. Can you verify?

ohler55:
LGTM other than one open question.

ext/oj/dump.c
KJTsanaktsidis: Ah, what I've done is:

* Copy the implementation of `rails_xss_friendly_size` into `rails_friendly_size`, except change the `rails_xss_friendly_chars` lookup array to the `rails_friendly_chars` lookup array
* Dup the implementation of  `case RailsXEsc` into `case RailsEsc`
* `rails_xss_friendly_size` and `rails_friendly_size` return the number of bytes required for the string if there are no chars with code point > 127, or _negative_ of that number if there are such chars; the sign bit is used to sneak an extra bit of information out of the function.

There's a lot of code in common between them, but I didn't feel I understood the essence of this well enough to implement a better factoring of it. I'm hoping you have some suggestions :)

My best idea is...

* Rename `rails_xss_friendly_size` to `size_t required_buffer_size_for_escaped_string(const uint8_t *str, size_t len, const char *cmap, bool *has_hi_out)`. Make it accept the cmap to use as a parameter, essentially, and pass `hi` as an explicit out-param rather than smuggling it out via the sign bit
* Call `required_buffer_size_for_escaped_string` in `case RailsXEsc` and `case RailsEsc` with different cmaps (`rails_xss_friendly_chars` vs `rails_friendly_chars`)

Does this sound reasonable to you?

test/activesupport6/encoding_test.rb
KJTsanaktsidis: The original activesupport test _also_ runs with both `escape_html_entities_in_json` set to both values: this is running the test cases multiple times with different parameters: https://github.com/rails/rails/blob/56bcc0abd3c9a6b09469e9428f6eea0dd77c2294/activesupport/test/json/encoding_test.rb#L22-L37

so AFAICT this test is now running exactly as intended in the upstream repo and is identifical? I think I _removed_ the point of difference?

ext/oj/dump.c
ohler55: It does.

ohler55 (approved):

ohler55:
Thanks for the work. I know I was a little picky. Maybe too much so, sorry.

KJTsanaktsidis:
> Thanks for the work. I know I was a little picky. Maybe too much so, sorry.

Not at all! Thanks for your attention on this.

> My best idea is...
> Rename rails_xss_friendly_size to size_t required_buffer_size_for_escaped_string(const uint8_t *str, size_t len, const char *cmap, bool *has_hi_out). Make it accept the cmap to use as a parameter, essentially, and pass hi as an explicit out-param rather than smuggling it out via the sign bit
> Call required_buffer_size_for_escaped_string in case RailsXEsc and case RailsEsc with different cmaps (rails_xss_friendly_chars vs rails_friendly_chars)

Do you want me to open another PR with those changes?

PR #915, opened 2024-02-28
CI: Add to dependency what to migrate to bundle gem in Ruby 3.4

Watson1978:
This patch will fix https://github.com/ohler55/oj/actions/runs/7908156404/job/21586745798

Some gems using CI will be changed bundled gem since Ruby 3.4.0. To use bundled gem, it requires to specify the dependency in Gemfile.

Ref. https://github.com/ohler55/oj/pull/901


ohler55 (approved):
Thank you

PR #921, opened 2024-05-10
Port to WASI by adding a check for getrlimit

kateinoigakukun:
This change adds WebAssembly/WASI target support for the extension library. Right now, we don't have an easy way to set up CI to test it, but we would appreciate it if we could merge the change into the upstream.

I manually tested it works well on my end.

### DIFF
--- a/ext/oj/extconf.rb
+++ b/ext/oj/extconf.rb
@@ -29,6 +29,7 @@ dflags = {
 have_func('rb_gc_mark_movable')
 have_func('stpcpy')
 have_func('pthread_mutex_init')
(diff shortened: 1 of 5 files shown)


ohler55:
Looks like the latest clang formatter has changed. I'm happy to merge once those issue are resolved. I can fix those tonight in the develop branch or if you are so inclined feel free to fix in your branch.

ohler55 (approved):

ohler55:
Thank you for fixing the clang formatting warnings.

kateinoigakukun:
Thanks!

PR #924, opened 2024-05-21
CI: Use `bundle exe` to run each test files in order to solve CI error

Watson1978:
Since Ruby-HEAD fails to load some gems, this patch gives `bundle exe` so that gems are loaded via bundler.

The following errors will be resolved:
image


ohler55 (approved):

ohler55:
Thank you for your continued support.

Issue #934, opened 2024-08-13
#as_json ignored on Rails 7.2.0


benedikt:
Reading through the other issues, I'm sorry to break the news, but it looks like your favorite issue with `#to_json` and `#as_json` is back with the recent release of Rails 7.2.0. The issue doesn't show up with Rails 7.1.x, but looking through the Rails source code it isn't immediately apparent what changed. 

The issue shows up upon calling `Oj.mimic_JSON`. Adding or removing `Oj.optimize_rails`, doesn't change the behavior. 

```ruby 
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'rails', '7.2.0' # No issues with 7.1.0
  gem 'oj', '3.16.5'
end

Oj.mimic_JSON # This seems to cause the issue
Oj.optimize_rails # Does not have an effect on the issue

class TestObject
  def as_json
    { foo: :bar }
  end
end

puts TestObject.new.to_json
```

Expected output: 

```
{"foo":"bar"}
```

Actual output: 

```
"#<Test:0x000000012024e9d0>"
```

Thanks a lot for this gem! I've been using it without any problems for _years_ now 💛

ohler55:
Sigh, Rails is a moving target. I'll look into it. Thanks for the test case. That helps a lot.

benedikt:
This change looks like the most likely culprit (at least to me), right now: https://github.com/rails/rails/commit/8e251a072815e24763787bccad3d69830e66bc90

Might be a red herring, though.

ohler55:
Looks like a likely cause. Just highlights the dangers of monkey patching. This will take some head scratching to find a work around.

benedikt:
I'm guessing the issue is caused by redefining `to_json` on `Object` in here? 
https://github.com/ohler55/oj/blob/f5e928a700a391815dc88e6ed25e8f04046b0b40/ext/oj/mimic_json.c#L908

Would it be an option to define a new module with the method on it and include it in `Object`? I think this is how the `json` gem does it as well. 

```ruby 
require 'json' 

Object.ancestors
# => [Object, JSON::Ext::Generator::GeneratorMethods::Object, PP::ObjectMixin, Kernel, BasicObject]
```

There would still be the problem of requiring everything in the right order, but I guess if `Oj.mimic_JSON` was added to application's `boot.rb` it would work? Alternatively, having a setup like this in the Gemfile would work as well.

```ruby 
[...]

PR #936, opened 2024-08-18
Do not override Object#to_json with mimic_json if ActiveSupport is existed

Watson1978:
Fix https://github.com/ohler55/oj/issues/934

### DIFF
--- a/test/activesupport7/abstract_unit.rb
+++ b/test/activesupport7/abstract_unit.rb
@@ -19,7 +19,10 @@ require "active_support"
 Thread.abort_on_exception = true
 
 # Show backtraces for deprecated behavior for quicker cleanup.
-ActiveSupport::Deprecation.debug = true
+if ActiveSupport::Deprecation.respond_to?(:debug)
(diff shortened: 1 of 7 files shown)


ohler55 (approved):

test/activesupport7/abstract_unit.rb
```diff
@@ -19,7 +19,10 @@
 Thread.abort_on_exception = true
 
 # Show backtraces for deprecated behavior for quicker cleanup.
-ActiveSupport::Deprecation.debug = true
+if ActiveSupport::Deprecation.respond_to?(:debug)
+  # Rails 7.2 does not have ActiveSupport::Deprecation.debug
```
HeyNonster, line 23: @ohler55 & @Watson1978 , for Rails 7.2 I think you want to call:

```ruby
ActiveSupport.deprecator.debug = true
```

Each module in Rails has its own deprecator instance now: https://github.com/rails/rails/commit/e5af9c298a108469a43758297ab56d12b3f0ddcf

PR #937, opened 2024-08-19
Update README.md

thromera:
Clarifications for the first example in the readme: 
- remove unneeded complex hash in comments 
- update the return value of the `puts` to match what actually would be displayed.


ohler55 (approved):

ohler55 (changes requested):
I'm glad to simply use `h == h2` but I prefer to include the output from Oj.dump so it's clear what the JSON would look like.

ohler55:
No response to requested changes.

PR #941, opened 2024-11-01
Properly declare global variables

byroot:
Fix: https://github.com/ruby/json/issues/672

GC compaction might move these objects around so the global C variable MUST be declared to the GC so it's updated there too.

### DIFF
--- a/ext/oj/mimic_json.c
+++ b/ext/oj/mimic_json.c
@@ -837,11 +837,15 @@ void oj_mimic_json_methods(VALUE json) {
     } else {
         json_error = rb_define_class_under(json, "JSONError", rb_eStandardError);
     }
+
+    rb_global_variable(&oj_json_parser_error_class);
     if (rb_const_defined_at(json, rb_intern("ParserError"))) {
         oj_json_parser_error_class = rb_const_get(json, rb_intern("ParserError"));
     } else {
         oj_json_parser_error_class = rb_define_class_under(json, "ParserError", json_error);
     }
+
+    rb_global_variable(&oj_json_generator_error_class);
     if (rb_const_defined_at(json, rb_intern("GeneratorError"))) {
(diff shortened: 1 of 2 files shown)


ext/oj/mimic_json.c
```diff
@@ -837,11 +837,15 @@ void oj_mimic_json_methods(VALUE json) {
     } else {
         json_error = rb_define_class_under(json, "JSONError", rb_eStandardError);
     }
+
+    rb_global_variable(&oj_json_parser_error_class);
```
ohler55, line 841: Shouldn't this be done after setting the value of the global?

byroot: No, it has to be done before:

```c
/**
 * Inform the garbage  collector that the global or static  variable pointed by
 * `valptr` stores  a live  Ruby object  that should not  be moved.   Note that
 * extensions  should use  this API  on  global constants  instead of  assuming
 * constants defined  in Ruby are  always alive.   Ruby code can  remove global
 * constants.
 *
 * Because this  registration itself has  a possibility  to trigger a  GC, this
 * function  must be  called  before any  GC-able objects  is  assigned to  the
 * address pointed by `valptr`.
 */
void rb_gc_register_address(VALUE *valptr);

/**
 * An alias for `rb_gc_register_address()`.
 */
void rb_global_variable(VALUE *);
```

ohler55: Thanks. LGTM then.

ohler55 (approved):

gmalette:
@ohler55 thanks for merging so fast! Can you release a new version please? The segfaults prevent us from upgrading

ohler55:
I will release this evening.

ohler55:
Released! Thanks for the fix.

PR #945, opened 2024-12-11
Fix streaming to file descriptor

### LINKED ISSUE #943: StreamWriter with subprocess - Undefined error: 0
Hi, thanks for oj - we've been using it for many years without problems, it's been rock solid.

We're now trying to pipe directly into stdin of a subprocess using `Oj::StreamWriter`, but we're running into I/O errors. I've put together a minimal test case here: https://github.com/jscheid/oj-subprocess-test

On macOS 14.7.1 with ruby 3.3.0 it consistently outputs the following:

```
$ bundle exec ruby test.rb
pushed chunk 0
pushed chunk 1
<SNIP>
pushed chunk 20
test.rb:18:in `push_json': Write failed. [_0_:Undefined error: 0] (IOError)
	from test.rb:18:in `block in <main>'
	from <internal:numeric>:237:in `times'
	from test.rb:17:in `<main>'
```

On Linux with ruby 3.3.1 the error is slightly different, here the message is `Write failed. [_0_:Success]`

With slightly different test data I've also witnessed (on macOS) `Write failed. [_25_:Inappropriate ioctl for device]`

With a larger `buffer_size` the error occurs later - it appears to occur just when the buffer is full. However, if I add `writer.flush` after every push, surprisingly it doesn't fail the first time `flush` is called.

All seems to work fine when writing the JSON data directly using `IO.write` calls, code for this is also included in the test case (toggle the if statement.) This leads me to believe that it isn't a problem with the `subprocess` gem (which we've also been using for years without issues.)

I've also tried simulating oj's buffering by writing it all with one single `IO.write` but that doesn't fail either.

Is anyone able to reproduce this? Is there perhaps some setting that we're missing or something else that's obvious?

jscheid:
Don't use write(2) for writing to file descriptors, use IO::write instead just as for writing to a non-file IO.

Using write(2) means having to handle all kinds of edge cases, such as EINTR and EAGAIN/EWOULDBLOCK, and doesn't release the GVL when blocking.

Fixes #943

### DIFF
--- a/ext/oj/stream_writer.c
+++ b/ext/oj/stream_writer.c
@@ -42,7 +42,8 @@ static void stream_writer_write(StreamWriter sw) {
 
     switch (sw->type) {
     case STRING_IO:
-    case STREAM_IO: {
+    case STREAM_IO:
+    case FILE_IO: {
         volatile VALUE rs = rb_str_new(sw->sw.out.buf, size);
 
         // Oddly enough, when pushing ASCII characters with UTF-8 encoding or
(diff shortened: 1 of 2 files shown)


ohler55:
If you can put in a conditional for the test with open3 Id be glad to merge you fix.

ohler55 (approved):

ohler55:
Released

PR #947, opened 2024-12-29
CI: Add Rails 8 and drop Rails 6.1

Watson1978:
- Rails 8.0
  - Ref. https://rubyonrails.org/2024/11/7/rails-8-no-paas-required
- Rails 6.1 has reached End of Maintenance
  - Ref. https://rubyonrails.org/2024/10/15/new-maintenance-policy-and-eol-annouments


ohler55 (approved):

ohler55:
Thank you.

PR #953, opened 2025-01-16
Add test for unecaped control characters in strings

byroot:
The JSON spec doesn't allow for unescaped control characters (`< 0x20`) in strings, but `Oj.load` does.

`Oj::Parser` correctly raises a parse error though.

### DIFF
--- a/test/test_parser_usual.rb
+++ b/test/test_parser_usual.rb
@@ -79,6 +79,16 @@ class UsualTest < Minitest::Test
     assert_equal({'ぴ' => '', 'ぴ ' => 'x', 'c' => 'ぴーたー', 'd' => ' ぴーたー '}, doc)
   end
 
+  def test_unescaped_ASCII_control_characters
+    p = Oj::Parser.new(:usual)
+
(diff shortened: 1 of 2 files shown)


ohler55:
This PR does not fix any existing issue but does cause all tests to break. If it can be made to fix the issue it can be reopened.

byroot:
This was a way to report the bug to you...

ohler55:
An issue is probably a better approach. If you can create one I'll address it. If you don't feel so inclined I create an issue for this. I'll use your tests as part of the description of the issue.

byroot:
Do what you want it's your gem. I thought I was doing a favor by providing a test case, but apparently not. Next time I'll keep the bugs I find for myself I guess.

ohler55:
Please don't take it negatively. It's just that I can't really work with a PR that causes builds to fail. Kind of breaks the development process. I am happy to accept an new issue but PRs are for a different purpose.

ohler55:
See https://github.com/ohler55/oj/issues/968

PR #954, opened 2025-01-23
Mention drop of Ruby 2.6 in OJ 3.14.3

brentkearney:
- To save developers on older versions of Rails some time in finding the latest OJ version compatible with Rails 5.2 / Ruby 2.6, mention the drop of support in the CHANGELOG.


ohler55:
Looks like Rails 7 has some issues not related to your change.

PR #956, opened 2025-02-06
github: add 3.4 for CI

kenhys:


Watson1978 (approved):
Hmm, it fails tests with Ruby HEAD.
We might need to fix it with another PR...

ohler55:
Better to fix in this PR since the only change is the addition of another ruby version.

ohler55:
Fair enough.

PR #973, opened 2025-06-11
Optimize `oj_dump_cstr` using SSE4.2 and SSSE3.

samyron:
### WORK IN PROGRESS - DO NOT MERGE

Hello! Me again.. this time optimizing `oj_dump_cstr` on x86-64 platforms using SSE4.2 and SSSE3. 

This PR is not yet ready for a thorough review but I wanted to get the discussion started which direction to the implementation.

There is still a bit of work to be done, particularly around reducing the duplication between the X86-64/AMD64 and ARM64 SIMD implementations. I haven't spent too much time trying to reduce this duplication. Additionally this only supports the `:compat` mode but `:rails` should be quick to support.

I noticed in the `oj/ext/extconf.rb` there is an option `--with-sse42` to compile SSE4.2 support. It looks like this is currently only used in the parser. This PR, however, currently uses runtime instruction set detection to determine if it can use the SSE4.2/SSSE3 functions. I use the [__builtin_cpu_supports](https://gcc.gnu.org/onlinedocs/gcc/x86-Built-in-Functions.html) function which is supported by GCC and clang to determine if the CPU supports the necessary instructions. I'm happy to continue down the path to support runtime CPU detection or switch to compile time support. The benefit of runtime CPU detection is anyone receiving a binary distribution of this library gets the SSE4.2 support so long as the platform compiling the library can compile SSE4.2/SSSE3 instructions (and currently using either GCC or clang). Additionally, consumers of the library don't need to provide any configure options to get the support. However, I'm not set on this approach. If you'd rather me switch this to an explicit compiler flag and/or use the existing `--with-sse42` flag, I'm happy to do so.

Additionally, even with runtime instruction set detection, I can provide a compiler flag to disable the feature (or enable.. depends if we want opt-in or opt-out).

## Benchmarks

CPU: Intel(R) Core(TM) i7-8850H

### Real world benchmarks
```
== Encoding activitypub.json (52595 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after     1.274k i/100ms
               after     12.759k (± 3.0%) i/s   (78.38 μs/i) -     64.974k in   5.097727s

Comparison:
              before:     8942.5 i/s
               after:    12758.8 i/s - 1.43x  faster


== Encoding citm_catalog.json (500298 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after    74.000 i/100ms
               after    740.405 (± 2.8%) i/s    (1.35 ms/i) -      3.700k in   5.001627s

Comparison:
              before:      760.2 i/s
               after:      740.4 i/s - same-ish: difference falls within error


== Encoding twitter.json (466906 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after   146.000 i/100ms
               after      1.478k (± 4.9%) i/s  (676.77 μs/i) -      7.446k in   5.054466s

Comparison:
              before:     1289.7 i/s
               after:     1477.6 i/s - 1.15x  faster


== Encoding ohai.json (20145 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after     2.016k i/100ms
               after     21.616k (± 4.4%) i/s   (46.26 μs/i) -    108.864k in   5.047241s

Comparison:
              before:    21127.7 i/s
               after:    21616.5 i/s - same-ish: difference falls within error
```

### Synthetic (happiest-of-happy paths - but still needs a single escape)

```
# benchmark_encoding "bytes.128.single-escape-at-end", ([(('a' * 127 + '\n'))] * 10000)

== Encoding bytes.128.single-escape-at-end (1330001 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after    63.000 i/100ms
               after    683.172 (± 3.1%) i/s    (1.46 ms/i) -      3.465k in   5.077817s

Comparison:
              before:      238.8 i/s
               after:      683.2 i/s - 2.86x  faster
```


samyron:
> Is this still a WIP?

Yes.. I at least need to clean up the warnings. Two questions for you though:

1. Are you okay you with runtime CPU detection to determine which method to use (on x86-64)? Or would you prefer using a compiler flag to enable/disable the feature?
2. If using runtime CPU detection,  would you like me to use that for the SSE4.2 code in the parser? If you would prefer the compile-time flag, would you like me to use the same `--with-sse42` flag for this code too?

samyron:
Apologies for the delays... it's been a busy week. I should be able to wrap this up within the next few days.

ohler55:
No worries. Same has happens to me on more than one occasion.

ohler55:
I was about ready to make a release. If your PR can be updated quickly I'll hold off for a bit.

samyron:
There is probably a clean-up ask to refactor the parser to use the `SIMD_Impl` initialized in `oj.c` (defined in `simd.h`).

PR #974, opened 2025-06-11
MAE-79380: Update to ruby 3.3.7

santiagogeliz:
https://vistahl.atlassian.net/browse/MAE-79380

Update oj gem to Ruby 3.3


ohler55:
I allowed this script to run before looking at the changes. I am rejecting this PR though as it is not correct to lock in a version of ruby for all users. Setting up a .ruby-version is very specific to an individual development or run time environment and should not be forced on all users.

PR #994, opened 2026-02-08
Add an ARM Neon string scanning implementation.

samyron:
This pull request adds an ARM Neon `string_scan` implementation to the parser.

## Benchmarks

Using the benchmarks from the `ruby/json` repository on my M1 Macbook Air:

```
== Parsing activitypub.json (58160 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after   772.000 i/100ms
               after      7.690k (± 0.7%) i/s  (130.04 μs/i) -     38.600k in   5.019888s

Comparison:
              before:     7043.9 i/s
               after:     7689.8 i/s - 1.09x  faster


== Parsing twitter.json (567916 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after    54.000 i/100ms
               after    543.495 (± 1.1%) i/s    (1.84 ms/i) -      2.754k in   5.067752s

Comparison:
              before:      519.4 i/s
               after:      543.5 i/s - 1.05x  faster


== Parsing citm_catalog.json (1727030 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after    27.000 i/100ms
               after    277.792 (± 1.8%) i/s    (3.60 ms/i) -      1.404k in   5.055628s

Comparison:
              before:      272.7 i/s
               after:      277.8 i/s - same-ish: difference falls within error


== Parsing ohai.json (32444 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after   867.000 i/100ms
               after      8.747k (± 1.2%) i/s  (114.32 μs/i) -     44.217k in   5.055570s

Comparison:
              before:     8412.1 i/s
               after:     8747.5 i/s - 1.04x  faster
```

### DIFF
--- a/ext/oj/simd.h
+++ b/ext/oj/simd.h
@@ -40,6 +40,7 @@ SIMD_Implementation oj_get_simd_implementation(void);
 // Count trailing zeros (for SSE2 mask scanning)
 #if defined(__GNUC__) || defined(__clang__)
 #define OJ_CTZ(x) __builtin_ctz(x)
+#define OJ_CTZ64(x) __builtin_ctzll(x)
 #elif defined(_MSC_VER)
 #include <intrin.h>
 static __inline int oj_ctz_msvc(unsigned int x) {
@@ -47,7 +48,15 @@ static __inline int oj_ctz_msvc(unsigned int x) {
     _BitScanForward(&index, x);
     return (int)index;
 }
+static __inline int oj_ctz64_msvc(uint64_t x) {
+    unsigned long index;
+    if (_BitScanForward64(&index, x)) {
+        return (int)index;
+    }
+    return 64;
+}
 #define OJ_CTZ(x) oj_ctz_msvc(x)
+#define OJ_CTZ64(x) oj_ctz64_msvc(x)
 #else
 // Fallback: naive implementation
 static inline int oj_ctz_fallback(unsigned int x) {
@@ -58,7 +67,17 @@ static inline int oj_ctz_fallback(unsigned int x) {
     }
     return count;
 }
+
+static inline int oj_ctz64_fallback(uint64_t x) {
+    int count = 0;
+    while ((x & 1) == 0 && count < 64) {
+        x >>= 1;
+        count++;
+    }
+    return count;
+}
 #define OJ_CTZ(x) oj_ctz_fallback(x)
+#define OJ_CTZ64(x) oj_ctz64_fallback(x)
 #endif
 
 // =============================================================================
--- a/ext/oj/simd.h
+++ b/ext/oj/simd.h
@@ -45,8 +45,12 @@ SIMD_Implementation oj_get_simd_implementation(void);
 #include <intrin.h>
 static __inline int oj_ctz_msvc(unsigned int x) {
     unsigned long index;
-    _BitScanForward(&index, x);
-    return (int)index;
+    if (0 == x) {
+        return 32;
+    } else {
+        _BitScanForward(&index, x);
+        return (int)index;
+    }
 }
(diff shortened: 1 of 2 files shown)


ext/oj/simd.h
```diff
@@ -40,14 +40,23 @@ SIMD_Implementation oj_get_simd_implementation(void);
 // Count trailing zeros (for SSE2 mask scanning)
 #if defined(__GNUC__) || defined(__clang__)
 #define OJ_CTZ(x) __builtin_ctz(x)
+#define OJ_CTZ64(x) __builtin_ctzll(x)
 #elif defined(_MSC_VER)
 #include <intrin.h>
 static __inline int oj_ctz_msvc(unsigned int x) {
     unsigned long index;
     _BitScanForward(&index, x);
```
samyron: This may be incorrect according to https://learn.microsoft.com/en-us/cpp/intrinsics/bitscanforward-bitscanforward64?view=msvc-170.

If `x == 0` the value written to `index` is undefined.

ohler55: Best to initialize when declared, no?

samyron: I'm not sure declaring the value would fix it as `_BitScanForward` is documented to have undefined behavior if `x == 0`. I think the fix is effectively: `(_BitScanForward(&index, x)) ? index : 32`.

That's how it is used in ruby's [ntz_int32](https://github.com/ruby/ruby/blob/0069009d35cf6b9ed78cf4fb5e5ce46dae21080f/internal/bits.h#L542) function.

I can create a separate PR to fix that as I didn't use that method in this PR and wanted to keep this limited to Neon support of the `string_scan` function.

samyron: Note: Looking at all of the `OJ_CTZ` callsites, each ensures that the `mask` is non-zero. While this may cause an issue if any future use calls this method with a `mask==0`, there aren't any issues today.

ohler55: If the default is 32 then how about just make line 46/47
```C
unsigned long index = 32;
```
Shouldn't add any overhead I think.

samyron: > If no bit is found, the function returns 0 and the value written to the address in the first parameter is undefined.

Even if we initialize it to `32` we don't have any sort of guarantee of what `_BitScanForward` will return. It's possible initializing `index` to `32` will work... though I think it will be a coincidence if it does as it's documented to be undefined.

ohler55: I think I see, so __BitScanForward when called with x of 0 sets index to an undefined value. Wouldn't a block like:

```c
if (0 == x) {
    index = 32;
} else {
    _BitScanForward(&index, x);
}
```

take care of that?

samyron: Yup, that will take care of it.

ohler55: If you can make that change I'll approve and merge.

ext/oj/simd.h
```diff
@@ -45,8 +45,12 @@ SIMD_Implementation oj_get_simd_implementation(void);
 #include <intrin.h>
 static __inline int oj_ctz_msvc(unsigned int x) {
     unsigned long index;
-    _BitScanForward(&index, x);
-    return (int)index;
+    if (0 == x) {
+        return 32;
+    } else {
```
ohler55: Minor nit, With the return on 49 there is no need for the `else`.

ohler55 (approved):

PR #997, opened 2026-03-10
Fix new parser stack consistency error

meinac:
The following JSON payloads cause "Stack inconsistency error", because the `result` function tries to return `Qundef` from the stack.

```ruby
parser = Oj::Parser.new(:usual)

parser.parse("{") # => !!!
parser.parse("[") # => !!!
```

With this change, we validate that the depth is always 0 after we finish parsing the JSON payload.

### DIFF
--- a/ext/oj/parser.c
+++ b/ext/oj/parser.c
@@ -1354,6 +1354,17 @@ static VALUE parser_missing(int argc, VALUE *argv, VALUE self) {
     return p->option(p, key, rv);
 }
 
+static void validate_final_depth(ojParser parser) {
+    if (0 == parser->depth) {
+        return;
+    }
+
+    switch (parser->stack[parser->depth]) {
+    case OBJECT_FUN: parse_error(parser, "Object is not closed");
+    case ARRAY_FUN: parse_error(parser, "Array is not closed");
+    }
+}
+
 /* Document-method: parse(json)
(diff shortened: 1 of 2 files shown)


ext/oj/parser.c
```diff
@@ -1354,6 +1354,16 @@ static VALUE parser_missing(int argc, VALUE *argv, VALUE self) {
     return p->option(p, key, rv);
 }
 
+static void validate_final_depth(ojParser parser) {
+    if (0 == parser->depth)
+        return;
```
ohler55:
```suggestion
    if (0 == parser->depth) {
        return;
    }
```

ohler55:
Thank you for identifying an issue and providing a fix.

ext/oj/parser.c
meinac: Thanks, @ohler55! I have applied this with my latest commit and squashed the commits. Would you mind giving this another look?

ohler55 (approved):

meinac:
Thanks for merging this fix @ohler55! Are you planning to release a new version with this fix soon?

ohler55:
Probably today. I did a little cleanup and am checking for other issues I can close relatively quickly.

ohler55:
Released

PR #998, opened 2026-03-14
Use fast_memcpy16 to quickly copy up to 16 bytes for tail handling in ARM Neon code and small copies when generating JSON.

samyron:
## Overview

This PR defines a `fast_memcpy16` function if `__builtin_memcpy` is available. This function compiles down to some branches, two loads and two store instructions which may overlap depending on the number of bytes to copy. The overlap will copy bytes in the same position so the bytes end up in the correct location. This function will get inlined into the caller and avoid function call overhead to `memcpy`. 

A similar [pull request](https://github.com/ruby/json/pull/924) was accepted to `ruby/json` with a very similar optimization and implementation.

I can update the SSE4 implementation to use the same optimization to this PR or as a follow up depending on your preference, assuming this PR is accepted.

## Benchmarks

These were run on an M1 Macbook Air. Before is the `develop` branch as of commit `c7fd9159ef285ba940a2075f51c382defe16a914`.

```
== Encoding activitypub.json (52595 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after     2.525k i/100ms
               after     25.255k (± 2.2%) i/s   (39.60 μs/i) -    126.250k in   5.001562s

Comparison:
              before:    22862.2 i/s
               after:    25254.9 i/s - 1.10x  faster


== Encoding citm_catalog.json (500298 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after   114.000 i/100ms
               after      1.142k (± 1.4%) i/s  (875.98 μs/i) -      5.814k in   5.093910s

Comparison:
              before:     1042.6 i/s
               after:     1141.6 i/s - 1.09x  faster


== Encoding twitter.json (466906 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after   246.000 i/100ms
               after      2.487k (± 1.6%) i/s  (402.10 μs/i) -     12.546k in   5.046023s

Comparison:
              before:     2217.1 i/s
               after:     2487.0 i/s - 1.12x  faster


== Encoding ohai.json (20145 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [arm64-darwin24]
               after     3.367k i/100ms
               after     33.844k (± 0.9%) i/s   (29.55 μs/i) -    171.717k in   5.074160s

Comparison:
              before:    30022.0 i/s
               after:    33844.5 i/s - 1.13x  faster
```

### DIFF
--- a/ext/oj/dump.c
+++ b/ext/oj/dump.c
@@ -257,7 +257,30 @@ inline static size_t hibit_friendly_size(const uint8_t *str, size_t len) {
         size += tmp;
     }
 
+#ifdef HAVE_FAST_MEMCPY
+    size_t total = 0;
+    if (len - i > 4) {
+        size += (len - i);
+        unsigned char buf[sizeof(uint8x16_t)];
+        memset(buf, ' ', sizeof(buf));
+        fast_memcpy16(buf, str, len - i);
+
+        uint8x16_t chunk  = vld1q_u8((const unsigned char *)buf);
+        uint8x16_t tmp1   = vqtbl4q_u8(hibit_friendly_chars_neon[0], chunk);
+        uint8x16_t tmp2   = vqtbl4q_u8(hibit_friendly_chars_neon[1], veorq_u8(chunk, vdupq_n_u8(0x40)));
+        uint8x16_t result = vorrq_u8(tmp1, tmp2);
+        uint8_t    tmp    = vaddvq_u8(result);
+
+        size += tmp;
+
+        total = size;
+    } else {
+        total = size + calculate_string_size(str, len - i, hibit_friendly_chars);
+    }
+#else
     size_t total = size + calculate_string_size(str, len - i, hibit_friendly_chars);
+#endif
+
     return total;
 #elif defined(HAVE_SIMD_SSE4_2)
     if (SIMD_Impl == SIMD_SSE42) {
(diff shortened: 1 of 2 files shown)


ext/oj/dump.c
```diff
@@ -257,7 +257,30 @@ inline static size_t hibit_friendly_size(const uint8_t *str, size_t len) {
         size += tmp;
     }
 
+#ifdef HAVE_FAST_MEMCPY
+    size_t total = 0;
+    if (len - i > 4) {
+        size += (len - i);
+        unsigned char buf[sizeof(uint8x16_t)];
+        memset(buf, ' ', sizeof(buf));
```
ohler55, line 265: Is this needed?

ohler55 (changes requested):
Just the one comment or question.

ext/oj/dump.c
samyron: In this case there are either fewer than 16 bytes left in `str` or `str` wasn't even 16 bytes long. We don't necessarily know if we can read past the end of `str`. In this case we create a temporary `buf` that is the vector width and ensure it's filled with characters that don't need to be escaped. We don't want anything past `len - i` to contribute to the size of the output string.

ohler55: If the intent is not have the characters contribute to the string length won't a null `\x00` be better since that is how C strings are terminated?

samyron: The issue here is we're looking at the entire 16 byte `uint8x16_t` vector, including those bytes past the end of `str`. That's why we create a `buf` with known characters in it. If we chose a `NULL` byte, those would each contribute an extra bytes towards the output. We'd have to subtract off `5 * (sizeof(uint8x16_t) - len - i)` from `size` as we aren't really going to output those `NULL` bytes.

Instead, we use a space as those output as a single byte. Additionally, the `hibit_friendly_chars_neon` table maps the _additional_ bytes required for each character. We we look up a space we get a `0` so they don't contribute any additional characters.

ohler55: That makes sense. Thanks for clarifying.

ohler55 (approved):

PR #1004, opened 2026-03-26
Fix read() not handling partial reads for large files

ursm:
## Summary

- `read()` system call may return fewer bytes than requested (POSIX allows short reads), especially for large files. Both `saj.c` and `parse.c` treated a short read as a failure, raising `IOError: failed to read from IO Object`. This change loops `read()` until all bytes are consumed.
- The same bug existed in two places: `oj_saj_parse` (saj.c) and `oj_pi_parse` (parse.c).

## Test plan

- Added `test_file` to `test/test_saj.rb` which passes a `File` object to `Oj.saj_parse`. This code path previously had no test coverage at all. The test verifies that parsing a File produces the same callbacks as parsing a String.
- Note: the test does not directly exercise the short-read fix, as triggering a partial `read()` on a regular file requires a file larger than ~2GB. The test serves as regression coverage for the File input path.

### DIFF
--- a/ext/oj/parse.c
+++ b/ext/oj/parse.c
@@ -1193,11 +1193,19 @@ oj_pi_parse(int argc, VALUE *argv, ParseInfo pi, char *json, size_t len, int yie
             buf      = OJ_R_ALLOC_N(char, len + 1);
             pi->json = buf;
             pi->end  = buf + len;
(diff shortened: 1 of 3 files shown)


ohler55 (approved):

ohler55:
Thank you for the fix.

PR #1005, opened 2026-04-08
Raise error for incomplete primitive literals

meinac:
Validate partially consumed literal tokens (`nul`, `fal`, `tru`) after parsing finishes, so the usual parser raises explicit errors instead of accepting invalid JSON.

This aligns primitive end-of-document checks with existing unclosed array/object validation and prevents silent success on malformed input.


ohler55 (approved):

ohler55:
thank you

PR #1007, opened 2026-04-14
Implement Oj::Parser.safe with configurable JSON safety limits

meinac:
Introduce a safe parser mode that enforces limits for array size, hash size, nesting depth, and total elements to reduce risk from oversized or deeply nested payloads.

### DIFF
--- a/ext/oj/safe.c
+++ b/ext/oj/safe.c
@@ -4,8 +4,9 @@ static VALUE max_hash_size_sym, max_array_size_sym, max_depth_sym, max_total_ele
     max_array_size_error_class, max_depth_error_class, max_total_elements_error_class;
 
 static void check_object_size(safe_T safe) {
-    if (NIL_P(safe->max_hash_size))
+    if (NIL_P(safe->max_hash_size)) {
         return;
+    }
 
     struct _usual usual                   = safe->usual;
     Col           current_object_location = usual.ctail - 1;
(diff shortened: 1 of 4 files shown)


ext/oj/safe.c
```diff
@@ -0,0 +1,223 @@
+#include "safe.h"
+
+static VALUE max_hash_size_sym, max_array_size_sym, max_depth_sym, max_total_elements_sym, max_hash_size_error_class,
+    max_array_size_error_class, max_depth_error_class, max_total_elements_error_class;
+
+static void check_object_size(safe_T safe) {
+    if (NIL_P(safe->max_hash_size))
```
ohler55: Please use `{` and `}` to bracket the if condition body.

ohler55:
I probably need to take a look at the clang-formatter rules. I'm surprised the conditionals without brackets didn't get flagged. Please add the brackets.

I didn't check all lines but it seemed like some might have been over 120 characters. If there are lines longer than that please change. If all are under 120 then it is fine.

The content, while different than my code looks fine otherwise. Thanks for the new addition.

meinac:
Thanks for your initial review @ohler55! I have changed the branches to always have brackets.

> I didn't check all lines but it seemed like some might have been over 120 characters. If there are lines longer than that please change. If all are under 120 then it is fine.

The longest line is exactly 120 characters, so it's fine I guess.

ohler55 (approved):

ohler55:
Thank you for the contribution.

PR #1009, opened 2026-05-11
Add `max_integer_digits` limit for legacy integer parsing

meinac:
Introduce a `max_integer_digits` option to cap parsed integer length in `Oj.load` and `Oj::Doc` legacy parsers.

This mitigates CPU DoS risk from untrusted JSON containing enormous integers while keeping default behavior unlimited.

### DIFF
--- a/ext/oj/fast.c
+++ b/ext/oj/fast.c
@@ -246,6 +246,19 @@ static void skip_comment(ParseInfo pi) {
 #define NUM_MAX (FIXNUM_MAX >> 8)
 #endif
 
+static void validate_integer_size(size_t limit, char *head, char *tail) {
+    size_t total       = (size_t)(tail - head);
+    bool   has_sign    = (head[0] == '-' || head[0] == '+');
+    size_t digit_count = total - (has_sign ? 1 : 0);
+
+    if (digit_count > limit) {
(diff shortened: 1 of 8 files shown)


ohler55:
Did you consider using the existing `int_range_max` and `int_range_min` field that already exist for output? That seems like the most direct approach but maybe you have some other thoughts on that.

meinac:
> Did you consider using the existing int_range_max and int_range_min field that already exist for output? That seems like the most direct approach but maybe you have some other thoughts on that.

@ohler55 - Good question, and yes, I have considered using them. There are two problems with those existing options;

1) They are for the JSON generation, so the users may want to have different configuration options for dumping vs parsing.
2) The real issue I'm trying to mitigate requires avoiding the [`rb_cstr_to_inum`](https://github.com/ohler55/oj/blob/babd7a11e78a221f6434c5e9e8b9ede9947243e3/ext/oj/parse.c#L999) call entirely because it's super slow and doesn't release the GVL while it's running, which is causing GVL contention. Therefore, the check has to happen even before we have a Ruby integer that can be used in comparison.

ohler55:
That makes sense.

ohler55 (approved):

PR #1021, opened 2026-06-25
Fix Set serialization/deserialization on Ruby 4

fable-stripe:
Ruby 4 made Set a built-in class (T_DATA) that no longer stores elements in an @hash instance variable. Oj's object mode relies on rb_ivar_foreach/rb_ivar_set for serialization, which produces an empty {"^o":"Set"} on Ruby 4 and fails to reconstruct on load.

This patch adds version-gated (RUBY_VERSION_MAJOR >= 4) handling:

Dump: detect Set objects with 0 ivars and emit the legacy format
  {"^o":"Set","hash":{"a":true,"b":true,...}}
  by iterating Set#to_a.

Load: when deserializing "hash" key on a Set, call Set#add for each
  hash key instead of rb_ivar_set. Also route Set (T_DATA on Ruby 4)
  through oj_set_obj_ivar instead of the error path.

Preserves full backward compatibility - on Ruby < 4 the existing ivar-based path is used unchanged.


ext/oj/dump_object.c
```diff
@@ -516,6 +516,26 @@ static void dump_obj_attrs(VALUE obj, VALUE clas, slot_t id, int depth, Out out)
     {
         int cnt = (int)rb_ivar_count(obj);
 
+        // Ruby 4: Set is a built-in class with no @hash ivar.
+        // Emit the legacy format: {"^o":"Set","hash":{"a":true,...}}
+        if (0 == cnt && RUBY_VERSION_MAJOR >= 4 &&
```
ohler55, line 521: Since the ruby version is known at compile time an `#if` would be a better way of conditionally using this code. Same with the other blocks in object.c.

Issue #961, opened 2025-04-17
Invalid dump/load result for Range object with same date at begin and end + circular option


artofhuman:
``` ruby
require 'date'
require 'oj'

date = Date.new(2021, 1, 1)
=> #<Date: 2021-01-01 ((2459216j,0s,0n),+0s,2299161j)>

period = date..date
=> #<Date: 2021-01-01 ((2459216j,0s,0n),+0s,2299161j)>..#<Date: 2021-01-01 ((2459216j,0s,0n),+0s,2299161j)>

str = Oj.dump(period, mode: :object, circular: true)
=> "{\"^u\":[\"Range\",{\"^O\":\"Date\",\"year\":2021,\"month\":1,\"day\":1,\"start\":2299161.0},\"^r1\",false]}"

Oj.load(str, mode: :object, circular: true)
=> #<Date: 2021-01-01 ((2459216j,0s,0n),+0s,2299161j)>..

Oj::VERSION
=> "3.16.10"

 
```

ohler55:
The "circular-range" branch should have a fix.

ohler55:
No response, closing.

PR #1038, opened 2026-07-14
Fix a shared Date or DateTime loading as nil with the circular option

### LINKED ISSUE #1036: Shared Date/DateTime reference (outside a Range) resolves to `nil` with `mode: :object, circular: true`
```ruby
require 'date'
require 'oj'

Oj::VERSION
# => "3.17.3"

date = Date.new(2026, 7, 6)
payload = {a: {x: date}, b: {x: date}}

dumped = Oj.dump(payload, mode: :object, circular: true)
# => {"^i":1,":a":{"^i":2,":x":{"^O":"Date","year":2026,"month":7,"day":6,"start":2299161.0}},":b":{"^i":4,":x":"^r3"}}

Oj.load(dumped, mode: :object, circular: true)
# => {:a=>{:x=>#<Date: 2026-07-06 ...>}, :b=>{:x=>nil}}   <-- expected the same Date, got nil
```

The dump never emits an `^i` reference id for the `Date` object itself (only the containing hashes get one), yet the second occurrence is still written as `"^r3"`  a backreference to an id that was never assigned. On load, that unresolvable reference silently becomes `nil` instead of raising or falling back to a full re-dump.

For comparison, every other type I tried round-trips correctly when shared the same way, including types that go through a similar "generic ivar" object path:

```ruby
require 'active_support/all'

tz = Time.current.in_time_zone('America/Toronto')
payload = {a: {x: tz}, b: {x: tz}}
dumped = Oj.dump(payload, mode: :object, circular: true)
Oj.load(dumped, mode: :object, circular: true)[:b][:x]
# => correct TimeWithZone value, NOT nil
```

`Time` (dumped via the compact `^t` format) and plain custom Ruby objects (dumped via the generic `^o` lowercase format) also round-trip correctly when shared only `Date`/`DateTime`'s special native fast-path (`^O` uppercase format) has this gap.

This looks related to #961 ("Invalid dump/load result for Range object with same date at begin and end + circular option"), which was fixed in 3.16.11 — I confirmed that specific `Range` case is fixed on 3.17.3. But this is a different, broader shape: a `Date` shared between two arbitrary sibling hashes, not nested inside a `Range`. It reproduces independently of `Range` entirely.

## Environment
- Oj 3.17.3
- Ruby 3.4 (also reproduces without Rails/ActiveSupport loaded — the `Date` case doesn't require ActiveSupport at all)

## Impact
In our case this silently corrupted production data, a `Date` shared by reference across multiple hashes in a larger payload (built by application code that reused one `Date` local variable across several hash literals) came back `nil` after a dump/load round trip with `circular: true`, with no error raised. We've since fixed our application code to avoid the aliasing, but wanted to flag the underlying gap since it's easy to hit unintentionally (a `date:` keyword-argument shorthand read from one local variable is a very natural pattern) and fails silently rather than raising.

Watson1978:
Fixes #1036.

## The bug

In the `:object` mode with `circular: true`, a `Date` (or `DateTime`) shared by two sibling hashes loads back as `nil`:

```ruby
date = Date.new(2026, 7, 6)
json = Oj.dump({a: {x: date}, b: {x: date}}, mode: :object, circular: true)
# {"^i":1,":a":{"^i":2,":x":{"^O":"Date","year":2026,"month":7,"day":6,"start":2299161.0}},":b":{"^i":4,":x":"^r3"}}
Oj.load(json, mode: :object, circular: true)
# {:a=>{:x=>#<Date: 2026-07-06 ...>}, :b=>{:x=>nil}}
```

Note the dump itself: the `Date` carries no `^i` id, yet the second occurrence is `"^r3"`, a back reference to an id that appears nowhere in the document. The id counter did advance, the top hash is `^i:1`, hash `a` is `^i:2`, and hash `b` is `^i:4`, so id `3` was consumed by the `Date` and never written. On load the unresolvable `^r3` becomes `nil` and the data is silently corrupted with no error.

## Root cause

`Date` and `DateTime` are registered as odd classes in `ext/oj/odd.c`. An odd class is dumped as `{"^O":class,...}` and is rebuilt with a create function when it is loaded, so there is no place to put a circular `^i` id on it and `dump_odd()` never writes one.

The problem is that the id is consumed before the odd branch is reached. `dump_data()` (`ext/oj/dump_object.c:49`) and `dump_obj()` (`ext/oj/dump_object.c:76`) both call `oj_check_circular()` first, which increments `out->circ_cnt` and registers the object in the circular cache, and only then call `dump_obj_attrs()`. The first thing `dump_obj_attrs()` does (`ext/oj/dump_object.c:470`) is to hand an odd class to `dump_odd()`, dropping the id it was given. The object is now in the circular cache with an id that was never written, so the next occurrence is dumped as a `^r` that refers to nothing.

Why the other types were not affected:

- `Time` is special cased in `dump_data()` before `oj_check_circular()` is called, so it never enters the circular cache and is dumped in full each time.
- `Rational` and `Complex` have their own dump functions that call `dump_obj_attrs()` with the id `0` and never call `oj_check_circular()`, so they too are dumped in full each time. `Rational` is an odd class and round trips correctly today for exactly this reason.
- A plain object dumped as `^o` reaches `dump_obj_attrs()`, does not match an odd class, and writes its `"^i":<id>` there.

So the odd `^O` path was the only path that consumed a circular id without ever materializing it.

## The fix

Keep odd classes out of the circular cache, the same way `Time`, `Rational`, and `Complex` already are. The check is done at the point where the odd class is knowable and before the id is consumed, in a small helper shared by both `dump_data()` and `dump_obj()` so that both the `T_DATA` and the `T_OBJECT` entry points are covered:

```c
static void dump_circular_obj(VALUE obj, VALUE clas, int depth, Out out) {
    if (NULL != oj_get_odd(clas)) {
        dump_obj_attrs(obj, clas, 0, depth, out);
    } else {
        long id = oj_check_circular(obj, out);

        if (0 <= id) {
            dump_obj_attrs(obj, clas, id, depth, out);
        }
    }
}
```

A shared odd object is now dumped in full each time it is encountered and no dangling `^r` is written:

```ruby
Oj.dump({a: {x: date}, b: {x: date}}, mode: :object, circular: true)
# {"^i":1,":a":{"^i":2,":x":{"^O":"Date",...}},":b":{"^i":3,":x":{"^O":"Date",...}}}
Oj.load(json, mode: :object, circular: true)
# {:a=>{:x=>#<Date: 2026-07-06 ...>}, :b=>{:x=>#<Date: 2026-07-06 ...>}}
```

### Trade off

The shared odd object comes back as an equal object rather than as the same object.

Preserving the identity is possible. It would mean writing an `^i` inside the `^O` hash and, on load, re-registering the instance in the circular array in `end_hash()` after the create function has run, since at parse time the value on the stack is still the class and not the instance yet. That needs a new field on the `Val` stack entry to carry the id from the `^i` key to the close of the hash, a new case in `hash_set_num()` for the `^i` key on a `T_CLASS` value, which today is fed to `oj_odd_set_arg()` and raises `^i is not an attribute of Date`, and a change to the dump side. It also changes the `^O` format, so a dump written by this version would raise a parse error on an older Oj.

That is a lot of surface, on both the dump and the load side, for an identity that `Date`, `DateTime`, `Rational`, and `Complex` do not need since they are immutable value types. Given that `Time`, `Rational`, and `Complex` are already dumped in full every time, doing the same for odd classes is both the smaller and the more consistent fix, so the identity preserving version is left aside for now. The `^o` path, where the identity is preserved, is untouched.

The attributes of an odd object are dumped normally, so a plain object or hash shared between two odd objects is still shared through `^i` and `^r` as before.

## Relation to #961

Issue #961, a `Range` with the same `Date` as the begin and the end, was worked around in 3.16.11 by disabling the circular option inside the `Range` struct dump. That is the same underlying problem seen through a different shape. #1036 needs no `Range` at all, only a `Date` shared between two arbitrary values. The `Range` workaround keeps working and is now covered by a test so it cannot regress.

## Tests

Added to `test/test_object.rb`:

- `test_circular_shared_date` and `test_circular_shared_datetime`, a `Date` and a `DateTime` shared by two sibling hashes, assert that no `^r` is written and that both values load back equal to the original.
- `test_circular_shared_date_in_range` and `test_circular_shared_datetime_in_range`, the #961 shape, as a regression guard.
- `test_circular_shared_odd`, a class registered with `Oj.register_odd` and shared, asserts the exact dump and that the hash the odd object is built from is still shared through `^i` and `^r`.

The three new tests for the shared `Date`, `DateTime`, and odd class fail on develop and pass with this change. `test/tests.rb` is green, 507 runs, 1258 assertions, 0 failures, 0 errors.

`rake test:valgrind` (ruby_memcheck) on Linux is clean, 484 runs, 1231 assertions, 0 failures, 0 errors, and no memory errors reported.

### DIFF
--- a/ext/oj/dump_object.c
+++ b/ext/oj/dump_object.c
@@ -10,6 +10,24 @@ static const char hex_chars[17] = "0123456789abcdef";
 
 static void dump_obj_attrs(VALUE obj, VALUE clas, slot_t id, int depth, Out out);
 
+// An odd class is dumped as {"^O":class,...} and is rebuilt with a create
+// function when loaded so there is no place to put a "^i" circular id on it.
+// Registering it in the circular cache would consume an id that is never
+// written and any later occurrence would then be dumped as a "^r" that refers
+// to nothing. Instead dump odd classes in full each time they are encountered
+// just like Time, Rational, and Complex are.
+static void dump_circular_obj(VALUE obj, VALUE clas, int depth, Out out) {
+    if (NULL != oj_get_odd(clas)) {
+        dump_obj_attrs(obj, clas, 0, depth, out);
+    } else {
+        long id = oj_check_circular(obj, out);
+
+        if (0 <= id) {
+            dump_obj_attrs(obj, clas, id, depth, out);
+        }
+    }
+}
+
 static void dump_time(VALUE obj, Out out) {
     switch (out->opts->time_format) {
     case RubyTime:
@@ -47,11 +65,7 @@ static void dump_data(VALUE obj, int depth, Out out, bool as_ok) {
                 oj_dump_cstr(str, len, 0, 0, out);
             }
         } else {
-            long id = oj_check_circular(obj, out);
-
-            if (0 <= id) {
-                dump_obj_attrs(obj, clas, id, depth, out);
-            }
+            dump_circular_obj(obj, clas, depth, out);
         }
     }
 }
@@ -74,11 +88,7 @@ static void dump_obj(VALUE obj, int depth, Out out, bool as_ok) {
             oj_dump_raw(str, len, out);
         }
     } else {
-        long id = oj_check_circular(obj, out);
-
-        if (0 <= id) {
-            dump_obj_attrs(obj, clas, id, depth, out);
-        }
+        dump_circular_obj(obj, clas, depth, out);
     }
 }
 
--- a/test/test_object.rb
+++ b/test/test_object.rb
@@ -207,6 +207,24 @@ class ObjectJuice < Minitest::Test
     end # Ni
   end # Ichi
 
+  class Oddball
+
+    attr_reader :h
+
+    def initialize(h)
+      @h = h
+    end
+
+    def self.create(h)
+      new(h)
+    end
+
+    def ==(other)
+      other.is_a?(Oddball) && other.h == @h
+    end
+
+  end # Oddball
+
   def setup
     @default_options = Oj.default_options
   end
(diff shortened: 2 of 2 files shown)


ohler55 (approved):

ohler55:
@Watson1978 you are awesome! Let me know when you are done and I'll make a release.

Watson1978:
@ohler55
I confirmed the TODO report from claude agent.
Now, there is no blocker for release 👍🏻 
[SECURITY_AUDIT-oj.en.md](https://github.com/user-attachments/files/30008520/SECURITY_AUDIT-oj.en.md)

ohler55:
Thanks

ohler55:
released

PR #1084, opened 2026-07-30
Fix rake never running the test/test_*.rb suite

Watson1978:
`test_all` invokes `Rake::Task['test']`, but the `Rake::TestTask` that defined it was commented out with `=begin` / `=end`. Rake therefore synthesized a **file task for the existing `test/` directory**, which has no actions, so the invoke was a no-op that neither ran anything nor complained:

```ruby
Rake::Task.task_defined?("test")  # => false
Rake::Task["test"].class          # => Rake::FileTask
Rake::Task["test"].actions.size   # => 0
```

## Consequence

`bundle exec rake` — the default task, and what CI runs — executed only `tests.rb`, `tests_mimic.rb`, `tests_mimic_addition.rb` and `test/json_gem`. **None of the `test/test_*.rb` files ran.**

Most of them appear in the memcheck list, which covers one platform and one Ruby. These ran nowhere at all:

`test_debian` · `test_generate` · `test_parser` · `test_parser_saj` · `test_realworld` · `test_scp` · `test_various`

| | runs | assertions |
|---|---|---|
| `bundle exec rake` before | 569 | 2,722 |
| `bundle exec rake` after | **1,312** | **14,496** |

## Why each file gets its own process

Every one of them passes on its own, but loading them all into a single process does not work. They share process-global state — `Oj.default_options`, the `Oj::Parser.saj` and `Oj::Parser.usual` singletons, `Oj.mimic_JSON` — so the result depends on collection order:

| seed | single shared process |
|---|---|
| 1 | 0 failures |
| 2 | 0 failures |
| 42 | 13 failures, 1 error |
| 1234 | 10 failures |
| 9999 | 9 failures |

(Saving and restoring `Oj.default_options` around every test cuts that from 43 problems to 7, so most of it is that one option set, but not all of it.)

This is presumably why the task was commented out in the first place. Isolating the shared state is worth doing, but it is a change to the tests rather than to how they are collected, so this PR runs each file in its own process — the way `test_all` already runs `test/json_gem/*_test.rb`. The glob deliberately does not reach `test/isolated/` (each of those needs its own process and setup), `test/json_gem/` (run separately by `test_all`) or the activesupport suites (they have their own tasks).

## One list for both tasks

`rake test` and `rake test:valgrind` now start from the same `TEST_FILES`, so a new `test/test_*.rb` is picked up by both without being added by hand anywhere. The memcheck list is that set minus two files, because `ruby_memcheck` loads everything it is given into one process:

* **`test_generate.rb`** calls `Oj.generate`, which calls `Oj::Rails.mimic_JSON` implicitly. After that a parse error is a `JSON::ParserError` rather than an `Oj::ParseError`, so the six tests in `test_max_integer_digits.rb` that name the class fail. Nothing undoes it, which is why the file belongs alongside `test/isolated/`.
* **`test_scp_fork.rb`** is new here, holding the two `Oj.sc_parse` tests that fork a writer for the pipe they read. Valgrind traces the child as well and both processes write to the one XML report, which leaves it truncated: `Nokogiri::XML::SyntaxError: Premature end of data in tag valgrindoutput`. Splitting those two out keeps the other 21 tests of `test_scp.rb` under the lane, and it also settles the interference with `test_various` that had kept the whole file out of it.

| memcheck lane | files | runs | assertions |
|---|---|---|---|
| before | 18 | 656 | 11,318 |
| after | **24** | **690** | **11,356** |

Green on seeds 1, 42 and 1234, at 121–140 s locally.

## A name collision was hiding 22 tests

`test_saj.rb` and `test_parser_saj.rb` both defined `SajTest`, so in a shared process whichever loaded second **discarded 22 of the other's tests**, with nothing but a `-w` warning to say so:

```
test/test_saj.rb:155: warning: method redefined; discarding old test_full
test/test_parser_saj.rb:236: warning: previous definition of test_full was here
```

The parser one is now `ParserSajTest`. Both files also defined an identical `AllSaj`, which moves to `test/saj_handlers.rb`, next to the `test/scp_handlers.rb` that the two scp files share. Neither is named `test_*.rb`, so neither is collected as a suite of its own. A scan of every `test/test_*.rb` finds no duplicate class names left.

## Two tests had rotted unnoticed

Nothing was running them, so nobody saw these.

**`test_parser_saj.rb` `test_file`** opened `'saj_test.json'` — a fixture that was never committed, so this test cannot ever have passed. Added the fixture its assertions describe (`[true,false]`) and made the path relative to `__dir__` instead of to the working directory.

**`test_debian.rb` `test_as_json_object_compat_hash_cached`** expects `:use_as_json` to be honoured in `:compat` mode. It is not:

```
mode=compat use_as_json=true  -> "\"#<Orange:0x...>\""    as_json not called
mode=rails  use_as_json=true  -> {"json_class":"Orange"}  as_json called
mode=custom use_as_json=true  -> {"json_class":"Orange"}  as_json called
```

In `:compat` neither `as_json` nor `to_json` is called and the object is dumped as its `to_s`, so `Oj.load` hands back a String rather than an `Orange`. Whether `:compat` should honour `:use_as_json` is a question about intended semantics, not about test plumbing, so the test is skipped with that evidence recorded in a comment rather than having its expectation quietly edited. **If `:compat` is supposed to honour it, this is a live bug and the skip should become a fix.**

## Verification

`rake test` is green over the 26 files, 743 runs and 11,774 assertions, and a deliberately added failing test makes it abort with a non-zero exit and name the file:

```
failed: test/test_zz_gate.rb
```

`bundle exec rake` end to end is green at 1,312 runs and 14,496 assertions. `rake test:valgrind` is green with no Valgrind report and no redefinition warnings.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,6 +9,13 @@ Rake::ExtensionTask.new('oj') do |ext|
   ext.lib_dir = 'lib/oj'
 end
 
+# The one list both `rake test` and `rake test:valgrind` start from, so a new
+# test/test_*.rb is picked up by both without having to be added by hand
+# anywhere. test/isolated/ is not matched: each of those needs its own process
+# with specific setup. Neither is test/json_gem/ (test_all runs it separately)
+# nor the activesupport suites (they have their own tasks).
+TEST_FILES = FileList['test/test_*.rb'].to_a.sort.freeze
+
 if RUBY_PLATFORM.include?('linux')
   begin
     require 'ruby_memcheck'
@@ -24,20 +31,6 @@ if RUBY_PLATFORM.include?('linux')
       valgrind_suppressions_dir: 'test/valgrind'
     )
 
-    # ruby_memcheck runs every listed file in a single process. Only the files
-    # that test/tests.rb loads are known to coexist that way; the other
-    # test_*.rb files mutate global state (Oj.default_options, mimic_JSON, ...)
-    # or fork, and upstream runs those in their own processes. Mirror the
-    # tests.rb set here, minus test_scp which forks and talks over a socket.
-    # test_parser_usual is not in tests.rb but coexists with the rest and is
-    # the only coverage of the Oj::Parser options.
-    memcheck_test_files = %w[
-      test_compat test_custom test_fast test_file test_gc test_hash
-      test_integer_range test_long_strings test_max_integer_digits test_null
-      test_object test_parser_safe test_parser_usual test_rails test_saj
-      test_strict test_wab test_writer
-    ].map { |name| "test/#{name}.rb" }
-
     namespace :test do
       task :check_valgrind do
         unless system('command -v valgrind > /dev/null 2>&1')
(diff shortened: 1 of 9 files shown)


Watson1978:
Marking this a draft: the Windows lanes need #1076 first.

Running these files on Windows for the first time surfaced a third thing that had been hidden by them never running. `test_parser_saj.rb`'s `test_big_value_file` and `test_big_key_file` fail there with

```
Errno::EACCES: Permission denied @ apply2files - D:/a/_temp/big20260730-9720-6m2zbv.json
```

Both wrap their body in `Tempfile.create(['big', '.json'])` and then call `p.file(f.path)`, which opens the same path a second time. `Oj::Parser#file` never closed its descriptor, so `Tempfile` could not delete a file that was still open. On Linux that is an invisible descriptor leak; on Windows deleting an open file is an error.

#1076 already fixes exactly that — it wraps the read loop in `rb_ensure` with a `close(fd)` — so this PR carries no change for it and is best merged after #1076.

For the record, this is not a regression from this PR: `develop`'s Windows lanes are green (run 30461093475), and the Linux and macOS lanes are green here too. Those two tests simply had never run on Windows before, because nothing was running `test/test_*.rb` at all.

ohler55 (approved):

ohler55 (approved):

PR #1091, opened 2026-07-31
Fix the 3.17.5 changelog entry saying only "Fixed multiple issues"

Watson1978:
The unreleased section reads

```markdown
## 3.17.5 - unreleased

- Fixed multiple issues.
```

which stands in for 47 merged pull requests. Someone deciding whether to take the upgrade, or working out whether a report they have already seen is handled, cannot tell any of that from one line. Most of the range is a security review of the C extension, so it is the kind of release where the detail is worth having.

This replaces the line with one bullet per pull request, grouped by what the change affects: memory safety, memory leaks, parsing and dumping, Rails, documentation, and tests and CI. Each bullet names the pull request it landed in, so the full write-up is one click away and the entry does not have to carry the mechanism.

The only commit in `v3.17.4..develop` with no bullet is #1060. It restored an `end` that was lost when #1058 and #1057 were merged into the same place in `test_object.rb`, so it never reached a release and there is nothing in it for a reader to act on.

Coverage was checked by extracting `(#N)` from both the new section and the commit subjects in the range: 48 commits, 47 bullets, the difference being #1060, with no duplicates and nothing referenced that is not in the range.

Documentation only. No code, no tests, no behaviour change.

`lib/oj/version.rb` still says `3.17.4` and the heading still says `unreleased`; bumping either belongs to release prep, not here.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,72 @@
 
 ## 3.17.5 - unreleased
 
-- Fixed multiple issues.
+Most of this release comes out of a security review of the C extension. There are no API changes.
+
+### Memory safety
+
+- Fixed hash flooding by seeding the key hash per process. The hash was decided entirely at compile time, so a document of colliding keys turned key interning into O(n^2) and left the process slow for every later parse. (#1053)
+- Fixed a one byte buffer over-read on a document that ends with a backslash inside an unterminated string. (#1054)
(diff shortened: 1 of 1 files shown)


ohler55 (approved):

ohler55:
I was going to have an AI generate from PRs before release but you just did. Thank you.

Watson1978:
I've fixed all the bugs found by Claude's large-scale code scan and filled in the changelog. 
Looking forward to the next release!

ohler55:
released

Watson1978:
Thanks 🎉

